PR #25, opened 2017-07-31
ErrorType names require a namespace prefix

uschi2000:


uschi2000:
@markelliot

markelliot:
* wonder if we should make namespace match the maven coords we use to identify a service
* should probably restrict use of `Default` to just this class

uschi2000:
maven coords is a bit problematic since they might change over time while APIs stay stable.
Default in this class: ACK

errors/src/main/java/com/palantir/remoting/api/errors/ErrorType.java
```diff
@@ -95,7 +103,22 @@ public static ErrorType create(Code code, String name) {
...
         }
-        return createInternal(code, name);
+        return createAndCheckNamespaceIsNotDefault(code, name);
+    }
+
+    private static ErrorType createAndCheckNamespaceIsNotDefault(Code code, String name) {
+        ErrorType error = createInternal(code, name);
+        Matcher matcher = ERROR_NAME_PATTERN.matcher(name);
+        if (!matcher.matches()) {
+            throw new IllegalStateException("Expected ERROR_NAME_PATTERN to match, this is a bug");
```
gracew: would probably be helpful to include the pattern and the name in the exception message. also, think this should be an IAE as `name` is user-specified in `create(Code, String)`

uschi2000: done https://github.com/palantir/http-remoting-api/pull/26

PR #28, opened 2017-08-04
Allow subclassing of ServiceException

uschi2000:


uschi2000:
simpler code + encapsulation of http-remoting API in fewer places.
```
throw new ServiceException(FOO_NOT_FOUND, SafeArg.of("foo", foo));
```
vs
```
throw new FooNotFoundException(foo);
```

uschi2000:
Thoughts?

markelliot:
In general I think we wan to encourage use of an Exception factory rather than simply extending + throwing exceptions. Not sure this makes a big difference if we're going to generate that factory?

markelliot:
also something to be said for knowing if the exception being thrown is in fact a ServiceException or just some other random exception

uschi2000:
I don't quite follow your two comments, Mark. I believe you're advocating for a `final` ServiceException in conjunction with something like
```
class MyAppErrors {
  static ServiceException accessDenied(...) { ...}
  ...
}
```
?

I don't really see a big difference here. Seems like subclassing is the canonical way to customize exceptions, but I don't really too mind either way. (Note that making ServiceException `final` does not mean that `MyAppError#myFooException` couldn't return a `FooException`, of course. I think we actually protect against this by registering exception mapping for ServiceException only in our app server; then throwing FooException is mostly worthless.)

markelliot:
With sub-classing, I’m worried that developers that join a code-base won’t recognize that not all exceptions are created equal. In particular, subclassing will make it less obvious (just by reading code) that there are such a thing as `ServiceException`s or that RandomCassandraException isn’t safe to rethrow for service errors while MySpecializedException is safe to rethrow.

If, instead, we create a factory to produce exceptions (like the one you describe), then when reading the code and seeing thrown exceptions it will be apparent/obvious that something is different.

As an alternative, we can allow sub-classing of exceptions but then need some tooling to, minimally, warn on non-built-in exceptions that don’t extend `ServiceException`. Because we’re in unchecked exception land, it might be hard to make that tooling catch enough of the problem cases.

uschi2000:
OK, sold.

PR #30, opened 2017-08-13
Add QosException

uschi2000:


errors/src/main/java/com/palantir/remoting/api/errors/QosException.java
```diff
@@ -0,0 +1,100 @@
...
+        return new Retry(Optional.empty(), Optional.of(backoff));
+
+    }
+
+    /**
+     * Returns a {@link Retry} exception indicating that the calling client should retry the request without delay and
+     * against the given node of this service.
+     */
+    public static Retry retryOther(URL redirectTo) {
+        return new Retry(Optional.of(redirectTo), Optional.empty());
```
schlosna: How do you envision cases where the other service is not directly accessible to the client (I.e. It clients requests are proxied through external load balancer and/or client only knows an external endpoint)?

uschi2000: Plan is to assume cooperate clients in most cases (for proxies: assumes that they forward the appropriate status codes and headers) and eventually add hard cut-offs in servers when they encounter non-cooperative clients.

uschi2000:
Corresponding change in http-remoting: https://github.com/palantir/http-remoting/pull/499

errors/src/main/java/com/palantir/remoting/api/errors/QosException.java
```diff
@@ -0,0 +1,100 @@
...
+ * service. Typically, this exception gets translated into appropriate error codes of the underlying transport layer,
+ * e.g., HTTP status codes 429, 503, etc. in the case of HTTP transport.
+ */
+public class QosException extends RuntimeException {
+
+    // Not meant for subclassing outside this class.
+    private QosException() {}
+
+    /**
+     * Returns a {@link Retry} exception indicating that the calling client should retry the request without delay and
```
bavardage: or... if you don't have a good idea for the duration backoff?? (I'm not convinced you always do, but unclear if `retryNow` and `retryLater` ends up as different codes (vs just adding a retry-after header).

I know I say this a lot, but I think we can't assume that remoting servers are only ever consumed by remoting clients - so having at least some hints as to how this gets turned into http is really useful.

errors/src/main/java/com/palantir/remoting/api/errors/QosException.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * An exception indicating that a request against this service may be retried, potentially against a different node
+     * of this service and with a given delay.
+     */
+    public static final class Retry extends QosException {
+        private final Optional<URL> redirectTo;
+        private final Optional<Duration> backoff;
+
+        private Retry(Optional<URL> redirectTo, Optional<Duration> backoff) {
```
bavardage: given we don't properly map both fields here being present, can we either
 - handle mapping that case
 - check here that they're not both set, throw IAE if they are

uschi2000:
As discussed, @bavardage : Uses subclasses now and removes the retryLater thing.

PR #33, opened 2017-08-18
failed precondition shouldn't be 500 error

ellisjoe:
returning a 500 for a failed preconditions seems to be the wrong http error code: https://tools.ietf.org/html/rfc7232#section-4.2


uschi2000:
In our context, a failed precondition is a server side error.

ellisjoe:
ah I missed the invalid argument type

PR #34, opened 2017-08-18
Use tokens, not tokens2

meditativeape:
@uschi2000 
Unfortunately, this is an API break and might require a major version bump...


uschi2000:
@pnepywoda @j-baker what was your plan with this? release 1.1 and hope that nobody runs into diamond dependencies?

pnepywoda:
Sure. There won't be diamond dependencies

PR #38, opened 2017-08-30
[BREAK] Rename QosException subclasses

uschi2000:


uschi2000:
This is an API break, but I'm willing to take the risk under the assumption that nobody has used this classes yet.

errors/src/main/java/com/palantir/remoting/api/errors/QosException.java
```diff
@@ -51,17 +51,16 @@ public static RetryOther retryOther(URL redirectTo) {
     }
 
     /**
-     * An exception indicating that all nodes of this service are currently unavailable and the client shall not attempt
-     * to wait for it to become available again. Typically, human intervention may be required to bring this service
-     * back up.
+     * An exception indicating that (this node of) this service are currently unavailable and the client may try again
```
chrisalice: "is currently"

uschi2000: Done here: https://github.com/palantir/http-remoting-api/pull/40

uschi2000: Thanks

PR #39, opened 2017-08-30
Add QosException.Visitor

uschi2000:


uschi2000:
FYI @gracew . I think this is uncontroversial, moving on.

PR #40, opened 2017-08-30
Add "Retry after" support to QosException.Throttle

uschi2000:


errors/src/main/java/com/palantir/remoting/api/errors/QosException.java
```diff
@@ -59,16 +69,24 @@ public static RetryOther retryOther(URL redirectTo) {
...
         return new Unavailable();
     }
 
     /** See {@link #throttle}. */
     public static final class Throttle extends QosException {
-        private Throttle() {}
+        private final Optional<Duration> retryAfter;
+
+        private Throttle(Optional<Duration> retryAfter) {
+            this.retryAfter = retryAfter;
```
chrisalice: Should this go into the exception message?

uschi2000: done

PR #45, opened 2017-09-06
compare toString on SafeArg

pnepywoda:


test-utils/src/main/java/com/palantir/remoting/api/testing/ServiceExceptionAssert.java
```diff
@@ -41,8 +44,15 @@ public final ServiceExceptionAssert hasType(ErrorType type) {
     public final ServiceExceptionAssert hasArgs(Arg<?>... args) {
         isNotNull();
 
-        if (!(actual.getArgs().equals(ImmutableList.copyOf(args)))) {
-            failWithMessage("Expected args to be %s, but found %s", Arrays.toString(args), actual.getArgs());
+        // toString is called on SafeArgs in SerializableError
+        List<Serializable> actualStrings = actual.getArgs().stream()
+                .map(arg -> arg instanceof SafeArg ? arg.toString() : arg)
```
meditativeape: Why only `SafeArg`? Both `SafeArg` and `UnsafeArg` share the same `toString` method.

test-utils/src/main/java/com/palantir/remoting/api/testing/ServiceExceptionAssert.java
```diff
@@ -41,8 +44,15 @@ public final ServiceExceptionAssert hasType(ErrorType type) {
...
+                .collect(Collectors.toList());
+        List<Serializable> givenStrings = Arrays.asList(args).stream()
+                .map(arg -> arg instanceof SafeArg ? arg.toString() : arg)
+                .collect(Collectors.toList());
+        if (!(actualStrings.equals(givenStrings))) {
```
meditativeape: Should also compare the Arg names -- Arg.toString() only contains value:
```
public final String toString() {
        return String.valueOf(value);
    }
```

test-utils/src/main/java/com/palantir/remoting/api/testing/ServiceExceptionAssert.java
pnepywoda: SerializableError only toString's SafeArg

uschi2000:
Don't quite grok the rationale for this?

pnepywoda:
CAn use Collections2.transform. Otherwise it's equals method doesn't work
vs a normal set or list to compare

uschi2000:
I see. Then I think the correct solution is to implement here whatever assertj does when you call something like `assertThat(someList).isEqualTo(someOtherList);`. I'm pretty they're not `toString()`ing.

pnepywoda:
The other reason is we tostring when logging. So yes I want to test it this
way anyway

uschi2000:
I don't think you're responsible for testing the exact mechanics of the logging behavior in your project.

pnepywoda:
It's actually "this" project not "my" project
https://github.com/palantir/http-remoting-api/blob/bb5ad9d248ce00aa275a1592fbcd11f618c38f91/errors/src/main/java/com/palantir/remoting/api/errors/SerializableError.java#L119

;)

test-utils/src/test/java/com/palantir/remoting/api/testing/ServiceExceptionAssertTest.java
```diff
@@ -18,10 +18,12 @@
 
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import com.google.common.collect.Collections2;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.remoting.api.errors.ErrorType;
 import com.palantir.remoting.api.errors.ServiceException;
+import org.assertj.core.util.Lists;
```
schlosna: realize this is test code, but probably want Guava's Lists

test-utils/src/main/java/com/palantir/remoting/api/testing/ServiceExceptionAssert.java
```diff
@@ -41,8 +39,30 @@ public final ServiceExceptionAssert hasType(ErrorType type) {
...
+            }
+            if (!expected.getName().equals(actual.getName())) {
+                failWithMessage("Expected arg %s to be named %s, but found %s", i, expected.getName(),
+                        actual.getName());
+            }
+            Object expectedValue =
+                    expected.isSafeForLogging() ? expected.getValue().toString() : expected.getValue();
+            Object actualValue =
+                    actual.isSafeForLogging() ? actual.getValue().toString() : actual.getValue();
+            if (!expectedValue.equals(actualValue)) {
```
schlosna: I believe `Arg` `value` is nullable, so need to handle `.getValue().toString()` and `expectedValue.equals` may NPE

PR #48, opened 2017-09-16
Add ServiceConfiguration maxNumRetries and backoffTimeSlot

uschi2000:


service-config/src/main/java/com/palantir/remoting/api/config/service/ServicesConfigBlock.java
```diff
@@ -79,6 +79,12 @@
     public abstract Optional<HumanReadableDuration> defaultWriteTimeout();
 
     /**
+     * Default global write timeout.
```
bavardage: this should probably match PartialServiceConfiguration.java#55

service-config/src/main/java/com/palantir/remoting/api/config/service/PartialServiceConfiguration.java
```diff
@@ -49,6 +49,15 @@
     /** Write timeout for requests. */
     Optional<HumanReadableDuration> writeTimeout();
 
+    /** The maximum number of times a failed RPC call should be retried. */
+    Optional<Integer> maxNumRetries();
+
+    /**
+     * The size of one backoff time slot for call retries. For example, an exponential backoff retry algorithm may
```
bavardage: (not saying javadoc should change, just curious) and so a 'constant' backoff could just retry every `backoffSlotSize`, a `linear` backoff choose `c*backoffSlotSize`?

uschi2000: Yes

service-config/src/main/java/com/palantir/remoting/api/config/service/ServicesConfigBlock.java
uschi2000: done

Issue #32, opened 2017-08-16
how is PartialServiceConfiguration class being used?


qinfchen:
ClientConfiguration can be created using ServiceConfiguration and there is no easy way to convert PartialServiceConfiguration to ServiceConfiguration or vice versa even though they are pretty much identical except for the absence of `security`.  I would like know more about how we intend to use the PartialServiceConfiguration class

qinfchen:
Can we make `ServiceConfigurationFactor.propagateDefaults` pubic?

uschi2000:
What's the use-case for this?

qinfchen:
Having a PartialServiceConfiguration declare outside the service discovery block like multipass'
```
auth:
  legacyModeEnabled: false
  multipass:
    security:
      trustStorePath: var/security/trustStore.jks
    uris:
    - https://multipass.palantir.dev:8643/multipass/api
```

or slate's postgate data source config
```
- dataSourcePoolCacheTimeout: 1m
  databaseName: palantir
  driver: org.postgresql.ds.PGSimpleDataSource
  maximumDataSourcePoolSize: 30
  name: Foundry
  portNumber: 5432
  postgate:
    uris: https://spp.palantir.dev:6473/postgate/api/
  serverName: postgres.palantir.dev
  type: foundry
```
It'd be great if we can specify service configuration outside of service discovery without requiring ssl block.  Unless we are encouraging teams to only take service configuration from service discovery?

uschi2000:
In order to use propagateDefaults, you'd need a handle on a ServiceConfigBlock; if you have one of those, why not use ServiceConfigurationFactory#get ?

qinfchen:
ServiceConfigBlock may not have the corresponding service config block for the PartialServiceConfiguration. I wonder if we should expose a convenient of converting PartialServiceConfiguration to a ServiceConfiguration by filling in the defaults that are not present, which is what propagateDefaults does mostly

uschi2000:
Such a method exists: build a custom ServiceConfigBlock and then call get().

meditativeape:
`ServiceConfigurationFactory#get` only works for services defined in the `ServicesConfigBlock`. In some places, we define partial service configurations outside of the service discovery block, for example workers, and event consumers. The defaults in `ServicesConfigBlock` should get propagated to those as well, but currently there's no good way of doing that.

ash211:
[...]

PR #54, opened 2017-10-04
add static `complete` method that takes partial and returns full config

qinfchen:


service-config/src/main/java/com/palantir/remoting/api/config/service/ServiceConfigurationFactory.java
```diff
@@ -35,6 +35,23 @@ public static ServiceConfigurationFactory of(ServicesConfigBlock services) {
         return new ServiceConfigurationFactory(services);
     }
 
+    public static ServiceConfiguration create(PartialServiceConfiguration partial) {
```
markelliot: it seems like a `ServiceConfigurationFactory#create` method ought to return a `ServiceConfigurationFactory`, rather than a `ServiceConfiguration`.

qinfchen: Any suggestions on the method name? since it is a factory class,  i thought using `create` to return a `ServiceConfiguration` would be appropriate. maybe be `get`?

ash211:
Fixes #32

service-config/src/main/java/com/palantir/remoting/api/config/service/ServiceConfigurationFactory.java
uschi2000: how about `complete`?

service-config/src/main/java/com/palantir/remoting/api/config/service/ServiceConfigurationFactory.java
```diff
@@ -35,6 +35,24 @@ public static ServiceConfigurationFactory of(ServicesConfigBlock services) {
         return new ServiceConfigurationFactory(services);
     }
 
+    /** returns a complete {@link ServiceConfiguration} for the given partial one. */
```
ash211: add Javadoc / `throws IllegalArgumentException` to signal to caller when / how this can fail

`@throws IllegalArgumentException when the given partial's security configuration is absent`

qinfchen:
any additional comments?

service-config/src/main/java/com/palantir/remoting/api/config/service/ServiceConfigurationFactory.java
```diff
@@ -35,6 +35,28 @@ public static ServiceConfigurationFactory of(ServicesConfigBlock services) {
         return new ServiceConfigurationFactory(services);
     }
 
+    /**
+     * returns a complete {@link ServiceConfiguration} for the given partial one.
+     *
+     * @throws IllegalArgumentException when the given partial's security configuration is absent
+     */
+    public static ServiceConfiguration complete(PartialServiceConfiguration partial) {
```
uschi2000: I'm a little scared of this method since it's very likely we're going to forget updating it when optional fields are added to ServiceConfiguration. Absent a proposal for making this compile-time or test-time safe, I'd prefer not taking this change at the expense of a little more consumer code (need to go through ServiceConfigBlock).

qinfchen: Good point. The current propagateDefaults method would have the same concern as mentioned. Since there is no compile-time or test-time safe, the consumer code will be prone to the same mistake, no?. Since partial and complete configurations only differ by the optionality of one property, would a common interface between them be better?

qinfchen: ping? I can help with whatever proposal we can come up.

uschi2000: Yes, the propagate method has the same problem, but at least it's only one place rather than two. We could remove or deprecate the builder and use full constructors, at least internally.

qinfchen: That works.  Could the pattern like below (from the Immutables doc) be another solution?

```
interface Vehicle {
}

interface VehicleBuilder {
  // Generated builders will implement this method
  // It is compatible with signature of generated builder methods where
  // return type is narrowed to Scooter or Automobile
  Vehicle build();
}

@Value.Immutable
public abstract class Scooter implements Vehicle {
  public abstract static class Builder implements VehicleBuilder {}
}

@Value.Immutable
public abstract class Automobile implements Vehicle {
  public abstract static class Builder implements VehicleBuilder {}
}

class Builders {
  void buildIt(VehicleBuilder builder) {
    Vehicle vehicle = builder.build();
  }

  void use() {
    buildIt(ImmutableScooter.builder());
    buildIt(ImmutableAutomobile.builder());
  }
}
```

uschi2000:
Have to check how this works with Jackson annotations and serialization

PR #57, opened 2017-11-07
Add mesh proxy type

jared2501:


service-config/src/main/java/com/palantir/remoting/api/config/service/ProxyConfiguration.java
```diff
@@ -37,10 +37,16 @@
...
         /**
-         * Use an http-proxy specified by {@link ProxyConfiguration#hostAndPort()}  and (optionally)
-         * {@link ProxyConfiguration#credentials()}.
+         * Use an http-proxy specified by {@link ProxyConfiguration#hostAndPort()}  and (optionally) {@link
+         * ProxyConfiguration#credentials()}.
          */
         HTTP,
+
+        /**
+         * Redirect requests to the {@link #hostAndPort} and sets the HTTP Host header to the original request's
```
uschi2000: "redirect" and "set" or "redirects" and "sets"

jared2501: done - chose "redirects"

PR #65, opened 2017-11-28
SerializableError#params now contains unsafe args in addition to safe args

iamdanfox:
Currently, products can throw `ServiceExceptions` containing both SafeArgs and UnsafeArgs, but only SafeArgs are included in the final JSON body that is returned from the server.

This PR changes this behaviour to send both safe and unsafe args over the wire.

### DIFF
--- a/errors/src/main/java/com/palantir/remoting/api/errors/SerializableError.java
+++ b/errors/src/main/java/com/palantir/remoting/api/errors/SerializableError.java
@@ -23,7 +23,6 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.palantir.logsafe.Arg;
 import java.io.Serializable;
 import java.util.Map;
-import java.util.stream.Collectors;
 import org.immutables.value.Value;
 
 /**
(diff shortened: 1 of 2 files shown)


errors/src/main/java/com/palantir/remoting/api/errors/SerializableError.java
```diff
@@ -107,20 +108,19 @@ String getMessage() {
 
     /**
      * Creates a {@link SerializableError} representation of this exception that derives from the error code and
-     * message, as well as the {@link Arg#isSafeForLogging safe} {@link ServiceException#args parameters}.
+     * message, as well as the {@link Arg#isSafeForLogging safe} and unsafe {@link ServiceException#args parameters}.
      */
     public static SerializableError forException(ServiceException exception) {
-        Builder builder = new Builder()
+        Map<String, String> safeAndUnsafeArgs = exception.getArgs()
+                .stream()
```
gracew: i think that a for-each was originally used b/c of perf concerns with streams

iamdanfox: TBH I really like the clarify of the toMap collectors - given that every arg is named by a real human dev in a YML API definition, I'd be surprised to ever see more than 10 args here

pkoenig10:
I think there needs to be a way to get only the safe parameters from a `SerializableError`. This is necessary for services to display error information to users while avoiding displaying unsafe information.

uschi2000:
When it makes it to the client then it can be displayed.

gracew:
I don't think a SerializableError is much different from any other types of responses, where we do not require annotation of safe vs unsafe information.

PR #71, opened 2018-03-06
Simplified assertions on ServiceException args, and more helpful error messages

mcintyret:
- args don't have to be in the 'correct' order since they are named in the assertions
- more helpful error messages that also print the name of the arg, not just its value


uschi2000:
cheers

PR #81, opened 2018-05-29
Add round robin request flag to PartialServiceConfiguration

justinbchan:
Provide mechanism to override current URL request handling (which pins to a healthy node). This allows for distributed traffic to downstream services (which may not be the case after a rolling restart of their nodes or if all instances of a given client randomly select the same URL).

### DIFF
--- a/service-config/src/main/java/com/palantir/remoting/api/config/service/PartialServiceConfiguration.java
+++ b/service-config/src/main/java/com/palantir/remoting/api/config/service/PartialServiceConfiguration.java
@@ -64,6 +64,11 @@ public interface PartialServiceConfiguration {
     /** Proxy configuration for connecting to the service. If absent, uses system proxy configuration. */
     Optional<ProxyConfiguration> proxyConfiguration();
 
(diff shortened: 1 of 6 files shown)


justinbchan:
Per offline discussion, I think a flag of some sort is necessary to override the current node-pinning functionality while the service mesh is not available.

uschi2000:
Yes, the real solution is service mesh, but I think we'll need this to unblock perf issues due to node-pinning.

markelliot:
How about we randomly reselect a node every 5 minutes?

________________________________

justinbchan:
Again, some sort of flag will still be necessary to enable this (non-default) behaviour.

The reselect-on-interval option may still result in clients loading up the same nodes for non-trivial periods of time. What's the downside of implementing a round-robin for each request?

uschi2000:
I think it'd be preferable to support two modes: failover-on-error, and round-robin (always).

justinbchan:
Per subsequent offline conversations, this option is going to manifest itself in code, not config

PR #83, opened 2018-06-05
switch to java-library-oss circle2 template

qinfchen:


build.gradle
```diff
@@ -6,18 +6,17 @@ buildscript {
 
     dependencies {
         classpath 'gradle.plugin.com.palantir:gradle-circle-style:1.1.2'
-        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.6'
-        classpath 'com.netflix.nebula:gradle-dependency-lock-plugin:4.2.0'
+        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.0'
         classpath 'com.netflix.nebula:nebula-dependency-recommender:5.2.0'
-        classpath 'com.netflix.nebula:nebula-publishing-plugin:5.1.4'
+        classpath 'com.netflix.nebula:nebula-publishing-plugin:7.2.2'
```
iamdanfox: I'd recommend doing this bump separately as it might contain breaking changes and is not strictly necessary to get us to Circle2 ;)

Publishing can be a real pain to get right!

build.gradle
```diff
@@ -62,24 +67,6 @@ subprojects {
 
     sourceCompatibility = 1.8
 
-    apply plugin: 'nebula.dependency-lock'
```
iamdanfox: I think locks might actually be appropriate for this repo... can we keep them for now?  We just need to make sure `verifyDependencyLocksAreCurrent` runs as part of `./gradlew build`

qinfchen: This is why i added `failOnVersionConflict` resolution strategy to force build failure when multiple version of deps are provided. Since we don't use range dependencies, I don't quite see benefits of using locks. 

Yeah, if we do plan to keep the locks, it'd be nice to use gradle locks.

uschi2000: I'd like to keep locks because I want to see in a PR when transitive dependencies change.

gradle/publish.gradle
```diff
@@ -1,9 +1,24 @@
-apply from: "${rootDir}/gradle/publish-base.gradle"
+apply plugin: 'com.jfrog.bintray'
```
iamdanfox: Please could we name this file `gradle/publish-jar.gradle` and use `name = rootProject.name` to make it 100% identical to the other conjure copies of this?  I'd quite like to write an excavator check that would just enforce exact consistency here :)

build.gradle
```diff
@@ -33,7 +33,7 @@ dependencies {
 }
 
 allprojects {
-    version System.env.CIRCLE_TAG ?: gitVersion()
+    version gitVersion()
```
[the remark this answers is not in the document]
qinfchen: it is kind of redundant,  https://github.com/palantir/http-remoting-api/pull/83/files#diff-1d37e48f9ceff6d8030570cd36286a61R40. or did i miss something?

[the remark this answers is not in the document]
uschi2000: the reason i had initially introduced this was to allow re-building with a new tag after a tagged build had failed.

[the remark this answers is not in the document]
uschi2000: maybe not a big deal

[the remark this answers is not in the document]
uschi2000: (because circle caches tags)

gradle/publish-jar.gradle
```diff
@@ -10,13 +10,15 @@ bintray {
     publish = true
     pkg {
         repo = 'releases'
-        name = 'http-remoting-api'
+        name = rootProject.name
```
iamdanfox: Gotta make sure you set this in `settings.grade` to be 'http-remoting-api' explicitly, otherwise gradle will try to infer it from the directory location (which will be wrong on CI because circle alwasy uses the `/root/project` path)

PR #92, opened 2018-06-21
Include ErrorType in ServiceException#getArgs

JacekLach:
Otherwise when a ServiceException is logged, it's type is only accessible via parsing the string message.

### DIFF
--- a/errors/src/test/java/com/palantir/remoting/api/errors/ServiceExceptionTest.java
+++ b/errors/src/test/java/com/palantir/remoting/api/errors/ServiceExceptionTest.java
@@ -18,9 +18,11 @@ package com.palantir.remoting.api.errors;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
+import java.util.List;
 import java.util.UUID;
 import org.junit.Test;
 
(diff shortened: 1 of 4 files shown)


JacekLach:
One gotcha this introduces: if I naively copy args for one service exceptions into another, it'll duplicate the `errorType` param:

```
ServiceException original = ...
ServiceException other = new ServiceException(NEW_ERROR_TYPE, original.getArgs());
```

This won't happen if you're constructing a service exception from a `SerializableError`.

Basic searches of internal codebases didn't find any instances of that pattern (it generally seems like a weird thing to do), so I think this is safe.

pkoenig10:
This creates bad UX for services that want to display error information because the error type will be duplicated.

I think we should create two types of `getArgs` methods to distinguish between:
- The args originally put in the service exception
- The args that should be logged (the implementation of the `SafeLoggable` method)

JacekLach:
Isn't that what getParameters is? It's public. Happy to rename if you have
a better concept.

pkoenig10:
Ah yes, sorry I missed that

errors/src/test/java/com/palantir/remoting/api/errors/ServiceExceptionTest.java
```diff
@@ -39,6 +41,32 @@ public void testExceptionMessageContainsNoArgs_safeLogMessageContainsSafeArgsOnl
...
+    @Test
+    public void testExceptionMessageWithNameCollisionWithInjectedArgs() {
+        Arg<?>[] args = {
+                SafeArg.of(ServiceException.ERROR_TYPE_ARG_NAME, "foo"),
+                UnsafeArg.of("arg2", 2),
+                UnsafeArg.of("arg3", null)};
+        ServiceException ex = new ServiceException(ERROR, args);
+
+        assertThat(ex.getLogMessage()).isEqualTo(EXPECTED_ERROR_MSG);
+        assertThat(ex.getMessage()).isEqualTo(EXPECTED_ERROR_MSG + ": {errorType=foo}");
```
ferozco: This seems undesirable. We should probably disallow usage or at the very least warn if there is a collision like this

JacekLach: I based that on the other test:

https://github.com/palantir/http-remoting-api/blob/da7dc5ff294b36e3ba8e6ef833ff5b812a7962e2/errors/src/test/java/com/palantir/remoting/api/errors/ServiceExceptionTest.java#L73-L76

It seems semantically service exception args is a set of `(key, value)` pairs with no collision detection; clients already have to be ready for collisions/duplicates. In part due to the fact that those come in over the network via a `SerializableError`, so when you're reading a `ServiceException` you can't trust that it doesn't have duplicate keys.

PR #100, opened 2018-07-23
http-remoting rebranding

qinfchen:
- [x] remove deprecations
- [x] rename things
- [x] move UserAgent from http-remoting (http-clients project) into here
- [x] rename group  to `com.palantir.conjure.java.api`
- [x] rename `com.palantir.remoting.api.*` package to `com.palantir.conjure.java.api.*`


service-config/build.gradle
```diff
@@ -4,7 +4,9 @@ apply from: "$rootDir/gradle/publish-jar.gradle"
 
 dependencies {
     compile project(":ssl-config")
+    compile "com.google.guava:guava"
```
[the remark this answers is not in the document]
markelliot: no

uschi2000:
Would it be easier to do this stuff in the new repository?

// mobile

________________________________

dansanduleac:
@uschi2000 we're actually going to keep this as a feature branch for the foreseeable future, and make develop provide easy conversions to the new types, so we can continue using both in parallel.
The reason we decided to do this in the same project is that we'd like to keep the project's history and issues, and all of this is mostly mechanic renames (removing only a fraction of things that are outside the scoep of conjure).

versions.props
```diff
@@ -1,6 +1,6 @@
 com.fasterxml.jackson.*:jackson-* = 2.6.7
 com.google.code.findbugs:jsr305 = 3.0.1
-com.google.guava:guava = 18.0
+com.palantir.safe-logging:* = 1.3.0
 com.palantir.safe-logging:safe-logging = 0.1.3
```
dansanduleac: we definitely don't want this one then :)

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
```diff
@@ -74,7 +71,7 @@ default UserAgent addAgent(Agent agent) {
     @Value.Check
     default void check() {
         if (nodeId().isPresent()) {
-            checkArgument(UserAgents.isValidNodeId(nodeId().get()),
+            Preconditions.checkArgument(UserAgents.isValidNodeId(nodeId().get()),
```
dansanduleac: remove this from this module, there's a `com.palantir.safe-logging:preconditions` package
https://github.com/palantir/safe-logging

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -148,13 +149,13 @@ private static UserAgent parseInternal(String userAgent, boolean lenient) {
...
 
     private static Map<String, String> parseComments(String commentsString) {
-        Map<String, String> comments = Maps.newHashMap();
-        for (String comment : COMMENT_SPLITTER.split(commentsString)) {
-            List<String> fields = COLON_SPLITTER.splitToList(comment);
-            if (fields.isEmpty()) {
+        Map<String, String> comments = new HashMap<>();
+        for (String comment : commentsString.split(",;")) {
+            String[] fields = comment.split(":");
+            if (fields.length == 0) {
```
dansanduleac: `fields` will actually always contain at least one item, even if `comment` is the empty string, so perhaps we wana check if `fields.length == 1 && fields[0].trim().isEmpty()`

qinfchen: we actually don't this check, updated the code to reflect the change.

PR #113, opened 2018-08-06
Failed precondition is 412

j-baker:


j-baker:
Ok. So is there a client side version of the error that'd be appropriate? in this case I want to throw it because the user's used up too many resources server side.

uschi2000:
Could look at adding 413, but you could also do CUSTOM_CLIENT, I guess

j-baker:
yeah, ok. As general feedback, given that this is the name of an actual HTTP response code, we probably should not have used it for one of our error codes with a different name

uschi2000:
Fair. I think copied these from gRPC’s names.

PR #115, opened 2018-08-23
[fix] Re-add SerializableError backcompat fields as write-only

dansanduleac:
This reverts the removals that occurred here: https://github.com/palantir/http-remoting-api/pull/100/files#diff-1604ffe60cf0f42a265fa29f94cbf660
And re-adds the fallback onto the old fields `exceptionClass` and `message` but in a cleaner way:
- no more weird exception message because of the cycles between `exceptionClass` <-> `errorCode` and `message` <-> `errorName`
- explicit, SafeLoggable errors when the required field as well as its fallback are both missing
- deprecated fallback fields no longer participate in equals, hashCode, toString; they are only used as a fallback if the original field wasn't set, otherwise they are essentially dropped

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
@@ -47,8 +47,6 @@ public abstract class SerializableError implements Serializable {
      * and/or name.
      */
     @JsonProperty("errorCode")
-    // TODO(rfink): errorCode delegates to exceptionClass if it's not set. This is quite a hack and should be removed
-    // when we remove support for the exceptionClass field.
     @Value.Default
     public String errorCode() {
         return getExceptionClass().orElseThrow(() -> new SafeIllegalStateException(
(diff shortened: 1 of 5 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
```diff
@@ -71,6 +85,26 @@ public String errorInstanceId() {
...
     public abstract Map<String, String> parameters();
 
+    /**
+     * @deprecated Used by the serialization-mechanism for back-compat only. Do not use.
+     */
+    @Deprecated
+    @JsonProperty(value = "exceptionClass", access = JsonProperty.Access.WRITE_ONLY)
+    @Value.Auxiliary
+    @SuppressWarnings("checkstyle:designforextension")
+    // TODO(rfink): Remove once all error producers have switched to errorCode.
```
iamdanfox: I don't think is really ever going to happen - possibly worth deleting entirely?

PR #117, opened 2018-09-03
[break] Rename to Conjure ecosystem

iamdanfox:
_We're rebranding http-remoting to be part of the 'conjure' ecosystem.  We've already tagged 2.0.0-rc7 on this branch so people have been trying it out successfully._

- maven group changed: `com.palantir.remoting-api` -> `com.palantir.conjure.java.api`
- packages changed:

	```diff
	+com.palantir.conjure.java.api.errors
	-com.palantir.remoting.api.errors
	```
- deleted deprecated code
- moved tracing classes to https://github.com/palantir/tracing-java
- new `com.palantir.conjure.java.api.config.service.UserAgent` class from http-remoting

## next steps

- tag 2.0.0 on this branch 
- rename the repository to `conjure-java-runtime-api`

## open questions

- ~should we just start using https://github.com/palantir/human-readable-types or does this expose us to more breaks than we want?~ minimal changes.
- ~should we do some kind of inheritance between old RemoteException and new RemoteException~ resolution: NO
- ~should we make a new API so that people can't pull out indexed safeArgs from SerializableErrors~ build this later


qinfchen:
> should we just start using https://github.com/palantir/human-readable-types or does this expose us to more breaks than we want?

we are already using https://github.com/palantir/http-remoting-api/blob/f212c60f193f3335f6586c195c57c935a94c3eac/service-config/src/main/java/com/palantir/remoting/api/config/service/HumanReadableDuration.java#L32, no?

> should we do some kind of inheritance between old RemoteException and new RemoteException?

Is the intention here to capture all try and catch of old RemoteException as if we would with new RemoteException? If we do that, wouldn't we run into the same issue in future when we remove the inheritance?

PR #118, opened 2018-09-04
More asConjure methods

iamdanfox:
This adds 

- [x] RemoteException#asConjure
- [x] SerializableError#asConjure
- [x] ErrorType#asConjure
- [x] ErrorType.Code#asConjure


errors/src/test/java/com/palantir/remoting/api/errors/SerializableErrorTest.java
```diff
@@ -71,13 +71,13 @@ public void forException_arg_key_collisions_just_use_the_last_one() {
     @Test
     public void testSerializationContainsRedundantParameters() throws Exception {
         assertThat(mapper.writeValueAsString(ERROR))
-                .isEqualTo("{\"errorCode\":\"code\",\"errorName\":\"name\",\"errorInstanceId\":\"\",\"parameters\":{},"
-                        + "\"exceptionClass\":\"code\",\"message\":\"name\"}");
+                .isEqualTo(
+                        "{\"errorCode\":\"code\",\"errorName\":\"name\",\"errorInstanceId\":\"\",\"parameters\":{}}");
```
qinfchen: this would be break, no?

iamdanfox: I think this test just captures exhaustive behaviour, but I don't think this is a real use case.  In practise, as long as you can deserialize both the old remoting2 and newer remoting3 JSON, I think it's fine.

PR #120, opened 2018-09-06
more asConjure for tracing 

qinfchen:
expose more asConjure for tracing in order to delegate all remoting tracer calls to `tracing-java`, so both remoting and new tracer share the same threadlocal.


markelliot:
it seems pretty broken that this would re-expose tracing -- the point of having a common and externalized tracing library is that everyone can use it.

iamdanfox:
This is no longer necessary - I've got the http-remoting migration PR implemented with a few package-private static utilities. https://github.com/palantir/http-remoting/pull/799/files#diff-64770984e33cab27b7b91ac92bc74daaR76

PR #130, opened 2018-09-26
Add JSON annotations to ErrorType

pkoenig10:
`ErrorType` is a safe-loggable arg of a `ServiceException`. Our internal logging library logs arg values using an object mapper. However `ErrorType` does not have JSON property annotations and consequently always gets serialized as "{}", which is not very helpful.

EDIT: Would prefer to merge https://github.com/palantir/conjure-java-runtime/pull/827 instead.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/ErrorType.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/ErrorType.java
@@ -16,6 +16,9 @@
 
 package com.palantir.conjure.java.api.errors;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/ErrorType.java
```diff
@@ -28,6 +31,9 @@
  */
 @Value.Immutable
 @ImmutablesStyle
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonSerialize(as = ImmutableErrorType.class)
+@JsonDeserialize(as = ImmutableErrorType.class)
```
uschi2000: do we need deserialization?

pkoenig10: Not that I know of, is there harm in allow it to be deserialized?

uschi2000: allow deserialization gives the impression that this is used as wire format, but that's not the case

pkoenig10:
Closing in favor of #131

PR #131, opened 2018-09-26
Log error name instead of error type

pkoenig10:
Follow up to #92 

`ErrorType` is a value type with multiple fields. Thus it is not an ideal arg value because it will be rendered as a JSON object, making it cumbersome to index, filter, read, etc.

Instead we should use the error name, which is simply a string. The error name is sufficient to fully determine the additional `ErrorType` information (each error name almost certainly has a single possible value for `code` and `httpCode`).

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
@@ -27,7 +27,7 @@ import javax.annotation.Nullable;
 
 /** A {@link ServiceException} thrown in server-side code to indicate server-side {@link ErrorType error states}. */
 public final class ServiceException extends RuntimeException implements SafeLoggable {
-    public static final String ERROR_TYPE_ARG_NAME = "errorType";
+    public static final String ERROR_NAME_ARG_NAME = "errorName";
 
     private final ErrorType errorType;
     private final List<Arg<?>> args;  // unmodifiable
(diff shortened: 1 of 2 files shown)


JacekLach:
:+1: this was actually my intention. Though I believe in ES etc you should be able to filter on subfields of the errortype? (i.e. 'errorType.name == MyException')?

errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
```diff
@@ -84,11 +84,11 @@ public String getLogMessage() {
     }
 
     /**
-     * Get Args of this service exception excluding injected/generated Args like errorType.
+     * Get Args of this service exception excluding injected/generated Args like errorName.
      * @return parameters passed to this ServiceException at construction time
      */
     public List<Arg<?>> getParameters() {
-        // errorType is always the first argument, so just slice it out
+        // errorName is always the first argument, so just slice it out
```
iamdanfox: I know this PR didn't introduce this, but I really don't like the fragile order reliance here, especially as I think these args come out of a JSON object (which is technically unordered!)

JacekLach: No, the ordering comes from how *we* build the list:

https://github.com/palantir/conjure-java-runtime-api/blob/bf7e7e8051d80c0dbc8fb7ad20db6b7bdb235bc9/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java#L123-L128

pkoenig10:
You may be able to filter by fields in ES, but all `ErrorType` are serialized as `{}`. This would be fixed by #130, but this seems like the better solution.

pkoenig10:
Closing in favor of #133

PR #133, opened 2018-09-28
Remove errorType from ServiceException args

pkoenig10:
Revert of #92 

Now that https://github.com/palantir/conjure-java-runtime/pull/827 has merged and we are logging the error name in the exception mapper, we no longer need the injected arg here.

With the injected args, the `ServiceExceptionMapper` logs would have both a `errorName` param and a `throwable0_errorType` param which is confusing and unnecessary.

Also note that this parameter has never been logged properly (see #130), so this does not cause a regression in capability.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
@@ -81,6 +81,16 @@ public final class ServiceException extends RuntimeException implements SafeLogg
         return args;
     }
 
+    /**
+     * Deprecated.
+     *
+     * @deprecated use {@link #getArgs}.
+     */
+    @Deprecated
+    public List<Arg<?>> getParameters() {
+        return getArgs();
+    }
+
     private static <T> List<T> copyToUnmodifiableList(T[] elements) {
         List<T> list = new ArrayList<>(elements.length);
         Collections.addAll(list, elements);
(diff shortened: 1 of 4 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
```diff
@@ -83,13 +81,10 @@ public String getLogMessage() {
         return args;
     }
 
-    /**
-     * Get Args of this service exception excluding injected/generated Args like errorType.
-     * @return parameters passed to this ServiceException at construction time
-     */
-    public List<Arg<?>> getParameters() {
```
uschi2000: was this released? is someone using this?

pkoenig10: This was released, but it is exactly the same as `getArgs`. But we probably need to keep it to avoid breaks. I'll keep it and mark it as deprecated.

uschi2000: not `probably`, but `absolutely`.

errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
```diff
@@ -27,7 +26,6 @@
 
 /** A {@link ServiceException} thrown in server-side code to indicate server-side {@link ErrorType error states}. */
 public final class ServiceException extends RuntimeException implements SafeLoggable {
-    public static final String ERROR_TYPE_ARG_NAME = "errorType";
```
dansanduleac: also an API break, but I don't think anyone should have relied on this so should be fine (and luckily test-utils didn't rely on it either)

PR #158, opened 2018-11-15
[fix][break] fix #148: ServiceException.getMessage includes unsafe args

### LINKED ISSUE #148: ServiceException rendered message should include both safe and unsafe parameters
The rendered exception message should only be used by loggers which do not support sls safe logging. In these cases, we should bias towards allowing debugging. Using sls logging unsafeParams are also recorded on disk, so this should not be considered dangerous.

This is a continuation of https://github.com/palantir/safe-logging/issues/76

cakofony:
This change provides parity with safe-logging Safe*Exception
types, see https://github.com/palantir/safe-logging/issues/76

By providing all argument data in the exception message we
provide full context to logging systems which do not understand
safe-logging types.


gatesn:
👍

PR #169, opened 2018-12-07
add fallback to CN verification option

tpetracca:


service-config/src/main/java/com/palantir/conjure/java/api/config/service/PartialServiceConfiguration.java
```diff
@@ -68,6 +68,10 @@
     @JsonAlias("enable-gcm-cipher-suites")
     Optional<Boolean> enableGcmCipherSuites();
 
+    /** Enables fallback to common name verification, defaults to false. */
+    @JsonAlias("fallback-to-common-name-verification")
+    Optional<Boolean> fallbackToCommonNameVerification();
```
j-baker: please make sure that this is deprecated?

j-baker: and note that it will be removed by the end of 2019?

PR #170, opened 2018-12-10
[improvement] Add SYSTEM proxy configuration

gatesn:
There is currently no way to defer to the JVM's `ProxySelector#getDefault` selector.


service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
```diff
@@ -36,6 +37,11 @@
 
     public enum Type {
 
+        /**
```
uschi2000: nit: `/** foo **/` (one line)

dansanduleac:
Closing as stale, please fill in the motivation according to the new template if you still need this (what is currently bad, what workflow is improved as a result).

PR #172, opened 2019-01-23
[improvement] UserAgents.tryParse precompiles the segment regex pattern

cakofony:


service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -110,9 +112,8 @@ public static UserAgent tryParse(String userAgent) {
     private static UserAgent parseInternal(String userAgent, boolean lenient) {
         ImmutableUserAgent.Builder builder = ImmutableUserAgent.builder();
 
-        Pattern segmentPattern = Pattern.compile(
```
iamdanfox: I wonder if we should write an error-prone check to suggest hoisting a `Pattern.compile` to a static field if it turns out to be compile time constant?

PR #182, opened 2019-02-11
Update docs. All args are propagated, not only safe ones

diogoholanda:
## Before this PR
Docs were misleading and said only safe args from a ServiceException were propagated

## After this PR
Docs are accurate and say all args are propagated


ferozco:
👍

PR #210, opened 2019-03-28
Add convenience method to retrieve the fully resolved service configuration

ellisjoe:

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServicesConfigBlock.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServicesConfigBlock.java
@@ -107,6 +107,13 @@ public abstract class ServicesConfigBlock {
     @JsonAlias("fallback-to-common-name-verification")
     public abstract Optional<Boolean> defaultFallbackToCommonNameVerification();
 
+    /**
+     * Returns the resolved {@link ServiceConfiguration} for the given service.
+     */
+    public final Optional<ServiceConfiguration> getServiceConfiguration(String serviceName) {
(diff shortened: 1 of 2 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServicesConfigBlock.java
```diff
@@ -107,6 +107,13 @@
     @JsonAlias("fallback-to-common-name-verification")
     public abstract Optional<Boolean> defaultFallbackToCommonNameVerification();
 
+    /**
+     * Returns the resolved {@link ServiceConfiguration} for the given service.
+     */
+    public final Optional<ServiceConfiguration> getServiceConfiguration(String serviceName) {
```
uschi2000: is this really necessary? not a huge fan of the circular dependency between config block and factory

uschi2000: (this class here is mostly a bean definition without much behavior)

ellisjoe: The thing that's sad is that you don't really want anyone to use this class and instead have everyone convert this directly into a `ServiceConfigurationFactory` (picking up the defaults here is hard otherwise). I'd honestly be pretty happy moving most of the factory into this class, and making this the singular api that you have to interact with as a consumer.

iamdanfox:
So apparently Jakub has other plans for how they're going to wire service config into their client library.  Let's close out for now - we can always revisit if necessary.

PR #227, opened 2019-04-12
[improvement] Make RemoteException implement SafeLoggable

gatesn:
Enable stacktraces containing remote exceptions to surface the error code and instanceId.

## Before this PR
Stack traces containing remote exceptions don't surface useful information like the instanceId and errorCode. 

## After this PR
Stack trace contains richer information by safe logging the exception message for remote exceptions.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -16,7 +16,9 @@
 
 package com.palantir.conjure.java.api.errors;
 
+import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
 import com.palantir.logsafe.UnsafeArg;
 import java.util.List;
@@ -50,14 +52,19 @@ public final class RemoteException extends RuntimeException implements SafeLogga
 
         this.error = error;
         this.status = status;
-        this.args = error.parameters().entrySet().stream()
-                .map(entry -> UnsafeArg.of(entry.getKey(), entry.getValue()))
-                .collect(Collectors.toList());
+        this.args = ImmutableList.<Arg<?>>builder()
+                .add(SafeArg.of("errorName", error.errorName()))
+                .add(SafeArg.of("errorCode", error.errorCode()))
+                .add(SafeArg.of("errorInstanceId", error.errorInstanceId()))
+                .addAll(error.parameters().entrySet().stream()
+                        .map(entry -> UnsafeArg.of(entry.getKey(), entry.getValue()))
+                        .collect(Collectors.toList()))
+                .build();
     }
 
     @Override
     public String getLogMessage() {
-        return getMessage();
+        return "RemoteException";
     }
 
     @Override
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -16,14 +16,21 @@
 
 package com.palantir.conjure.java.api.errors;
 
+import com.palantir.logsafe.Arg;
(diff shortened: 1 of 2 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -43,5 +50,18 @@ public RemoteException(SerializableError error, int status) {
 
         this.error = error;
         this.status = status;
+        this.args = error.parameters().entrySet().stream()
+                .map(entry -> UnsafeArg.of(entry.getKey(), entry.getValue()))
```
carterkozak: I'm less interested in the unsafe parameters from the SerializableError than the error-code, instance id, error name, and status code. Can we add those as safe args?

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -43,5 +50,18 @@ public RemoteException(SerializableError error, int status) {
...
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public String getLogMessage() {
+        return getMessage();
```
carterkozak: We can use a string constant for the log message, and provide everything else in `args`.

pkoenig10:
Can we keep this consistent with `ServiceException`? There we put the error type information in the log message and the only parameters are the given parameters.

carterkozak:
+1. We don't need to log the RemoteException unsafe params because they've already been logged by the service that originally threw the ServiceException.

pkoenig10:
That's actually specifically what I am advocating for. That is consistent with the behavior of `ServiceException`.

carterkozak:
Is it not feasible to look up the service exception as logged by the upstream service using the error-instance-id?

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
gatesn: The whole point is to get this into the stack trace. I don't think args (safe or unsafe) get printed out as part of that right?

pkoenig10:
It is. All I'm advocating for here is that the safe-logging behavior of `RemoteException` and `ServiceException` are consistent. If we want to change the behavior of `ServiceException` that doesn't seem unreasonable.

But we specifically removed the error type args from `ServiceException` in #133 because they were confusingly double logged by the `ServiceExceptionMapper`. You would also have to be careful about collisions between the given args and the injected error type args.

Although maybe it does make sense for the behavior to not be the same? Specifically, `ServiceException` should have only the given args and `RemoteException` should have only the error type args. But then you have the double logging problem in `RemoteExceptionMapper`.

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
carterkozak: Our safe-logging aware layouts include parameters in params/unsafeParams from the top level as well as cause and suppressed-cause exceptions using the parameter name prefix `throwable<INDEX>_`, for example `throwable_2_errorName`.

layouts which are not aware of safe-logging already include the full human-readable message in stack traces.

carterkozak:
Ahh, thanks for the clarification @pkoenig10!

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -43,5 +52,23 @@ public RemoteException(SerializableError error, int status) {
 
         this.error = error;
         this.status = status;
+        this.args = ImmutableList.<Arg<?>>builder()
```
pkoenig10: As I previously mentioned, I feel pretty strongly that this should be consistent with the implementation of `ServiceException`. `ServiceException` does not include user-provided args in it's safe-logging args for reasons described in #133.

I'm not opposed to changing this behavior, but if we want to combine error and user-provided args, we should do it for both `ServiceException` and `RemoteException`.

PR #236, opened 2019-04-21
[improvement] Fix safe-loggable exception errorprone warnings

carterkozak:
## Before this PR
Several validation failures resulted in unsafe exceptions which do not provide any information in some environments.

## After this PR
Safe exception variants are thrown.

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.java.api.config.service;
 
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -139,11 +140,14 @@ public final class UserAgents {
 
         if (!foundFirst) {
             if (lenient) {
-                log.debug("Invalid user agent, falling back to default/unknown agent",
-                        SafeArg.of("userAgent", userAgent));
+                if (log.isDebugEnabled()) {
+                    log.debug("Invalid user agent '{}', falling back to default/unknown agent",
+                            SafeArg.of("userAgent", userAgent));
+                }
                 return builder.primary(UserAgent.Agent.of("unknown", UserAgent.Agent.DEFAULT_VERSION)).build();
             } else {
-                throw new IllegalArgumentException("Failed to parse user agent string: " + userAgent);
+                throw new SafeIllegalArgumentException("Failed to parse user agent string",
+                        SafeArg.of("userAgent", userAgent));
             }
         }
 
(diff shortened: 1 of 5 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -176,7 +180,7 @@ static boolean isValidVersion(String version) {
             return true;
         }
 
-        log.warn("Encountered invalid user agent version", SafeArg.of("version", version));
+        log.warn("Encountered invalid user agent version '{}'", SafeArg.of("version", version));
```
iamdanfox: These `'{}'` placeholders are only really necessary if someone is not using sls-logging right? They make logs a tiny bit more ugly in our central log browsing tool, especially when you histogram by log message, so would prefer to omit them if that's reasonable

iamdanfox:
Thanks Carter! :+1:

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -139,11 +140,14 @@ private static UserAgent parseInternal(String userAgent, boolean lenient) {
...
-                        SafeArg.of("userAgent", userAgent));
+                if (log.isDebugEnabled()) {
+                    log.debug("Invalid user agent '{}', falling back to default/unknown agent",
+                            SafeArg.of("userAgent", userAgent));
+                }
                 return builder.primary(UserAgent.Agent.of("unknown", UserAgent.Agent.DEFAULT_VERSION)).build();
             } else {
-                throw new IllegalArgumentException("Failed to parse user agent string: " + userAgent);
+                throw new SafeIllegalArgumentException("Failed to parse user agent string",
+                        SafeArg.of("userAgent", userAgent));
```
iamdanfox: Given that we don't know the structure of this `userAgent` string, perhaps we should actually make this an `UnsafeArg`?

carterkozak: It was already logged as safe a couple lines above. I think it's a generally safe thing to log since it describes the client implementation, not the underlying data the client is operating on.

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
carterkozak: My preference is for our OSS shared tooling to support general-purpose loggers, especially since we don't provide an OSS sls logging tool.

iamdanfox: Eh, we're pushing for sls-logging everywhere internally, and the sls-logging [conjure defs](https://github.com/palantir/witchcraft-api/blob/master/witchcraft-logging-api/src/main/conjure/witchcraft-logging-api.yml) are open source already.  Could we take this as just another incentive to push slslog and our actual logback/log4j2 implementations out?

carterkozak: Works for me :-)

I'm not sure if that will cause problems for gradle plugins which use these libraries, since parameters will be dropped.

PR #242, opened 2019-05-07
Allow qos config per service

hsaraogi:
## Before this PR
Cannot set QoS configurations in config per service.

## After this PR
Can set QoS configurations in config per service. We often need to cut releases to modify the node selection strategy and failedCoolDown period. This allows us to get out of P0s with a simple config change.

## Possible downsides?
Too configurable.

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServiceConfiguration.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServiceConfiguration.java
@@ -44,6 +44,10 @@ public interface ServiceConfiguration {
 
     Optional<Integer> maxNumRetries();
 
+    Optional<NodeSelectionStrategy> nodeSelectionStrategy();
+
+    Optional<Duration> failedUrlCooldown();
+
     Optional<Duration> backoffSlotSize();
 
     Optional<Boolean> enableGcmCipherSuites();
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/NodeSelectionStrategy.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/NodeSelectionStrategy.java
@@ -1,0 +1,37 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.api.config.service;
+
+public enum NodeSelectionStrategy {
+
+    /**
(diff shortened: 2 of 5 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServiceConfiguration.java
```diff
@@ -44,8 +44,25 @@
...
+    /**
+     * Indicates how the target node is selected for a given request.
+     */
+    Optional<NodeSelectionStrategy> nodeSelectionStrategy();
+
+    /**
+     * The amount of time a URL marked as failed should be avoided for subsequent calls. If the
+     * {@link #nodeSelectionStrategy} is ROUND_ROBIN, this must be a positive period of time.
+     */
+    Optional<Duration> failedUrlCooldown();
```
pkoenig10: We probably don't want to add anything other than `failedUrlCooldown`. I don't think there is really a use case for making the node selection strategy and especially the QoS settings configurable.

hsaraogi: Some service where we depend on cache locality might need PIN_UNTIL_ERROR etc, so making node selection strategy configurable is a good idea. 

Agree that ClientQoS and ServerQoS is probably not required to be configured that often.

hsaraogi:
Why, we might want to change the failedUrlCooldown for services if a node is acting bad at runtime? https://support-jira.palantir.tech/browse/PDS-89144 could have been prevented with the change.

service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServiceConfiguration.java
pkoenig10: Does this need to be live-reloadable though? Whether a particular client uses PIN_UNTIL_ERROR or ROUND_ROBIN is probably something that should be set in code.

markelliot:
Why does failedUrlCooldown vary between installations / why should this be an 'operator' decision?

pkoenig10:
Having an failedUrlCooldown that is shorter than the socket timeout can result in a large number of requests hitting an unhealthy node. As an example, suppose the failedUrlCooldown is 10s and the socket timeout is 60s. This means than for 60s every 70s, 1/N requests are hitting a node that is unhealthy. This can very quickly lead to poor performance.

Currently the failedUrlCooldown is set in code. This meaning that quickly resolving these issues, requires making the change, cutting a new release, and getting that new release installed. If the failedUrlCooldown is configurable we would be able to remediate these issues much faster.

service-config/src/main/java/com/palantir/conjure/java/api/config/service/ServiceConfiguration.java
hsaraogi: will remove the selection strategy.

hsaraogi:
Why does failedUrlCooldown vary between installations / why should this be an 'operator' decision?

Thinking about it again, it is not. The problem was hit as build2 picked up the cooldown time of 10s which was lower than the idle timeout causing the issue. 

Say we have 3 nodes of service A and service B, and one node of service B is in a state that it is accepting connections but not processing requests, all requests will timeout after 60s which is the default OkHttp IDLE_TIMEOUT. 

So for the first 60s, serviceA sends 1/3rd requests to all three nodes, then one node is blacklisted for 10s and again build2 starts sending 1/3rd requests to all nodes. Thus, a lot of the requests failed. If the cooldown time is set to a higher value, like 5mins, this would be better.

hsaraogi:
closing in favor of https://github.com/palantir/conjure-java-runtime/pull/1098.

PR #243, opened 2019-05-08
Make RemoteException SafeLoggable and add parameters to message

pkoenig10:
## Before this PR
The implementations of `RemoteException` and `ServiceException` are not consistent:
- `RemoteException` does not implement `SafeLoggable`, `ServiceException` does
- `RemoteException#getMessage` does not include the parameters, `ServiceException#getMessage` does

## After this PR
The implementations of `RemoteException` and `ServiceException` are more consistent:
- Both implement `SafeLoggable`. `getLogMessage` does not include any args/parameters.
- `getMessage` includes all args/parameters, including unsafe args.

This is slightly different than #227 because this PR adds the parameters to the `RemoteException` message.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -16,14 +16,35 @@
 
 package com.palantir.conjure.java.api.errors;
 
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.UnsafeArg;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
 /**
  * An exception thrown by an RPC client to indicate remote/server-side failure.
  */
-public final class RemoteException extends RuntimeException {
+public final class RemoteException extends RuntimeException implements SafeLoggable {
     private static final long serialVersionUID = 1L;
 
     private final SerializableError error;
     private final int status;
+
+    private final String noArgsMessage;
+    private final List<Arg<?>> args; // unmodifiable
+
+    public RemoteException(SerializableError error, int status) {
+        super(renderUnsafeMessage(error));
+
+        this.error = error;
+        this.status = status;
+        this.noArgsMessage = renderNoArgsMessage(error);
+        this.args = copyArgsToUnmodifiableList(error);
+    }
 
     /** Returns the error thrown by a remote process which caused an RPC call to fail. */
     public SerializableError getError() {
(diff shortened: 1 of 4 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -35,13 +56,54 @@ public int getStatus() {
         return status;
     }
 
-    public RemoteException(SerializableError error, int status) {
-        super(error.errorCode().equals(error.errorName())
+    @Override
+    public String getLogMessage() {
+        return noArgsMessage;
```
dansanduleac: This shouldn't have things interpolated in it.
Can we add errorCode, errorName, errorInstanceId to the list of arguments instead?

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -35,13 +56,54 @@ public int getStatus() {
...
+    }
+
+    private static List<Arg<?>> copyArgsToUnmodifiableList(SerializableError error) {
+        Map<String, String> parameters = error.parameters();
+
+        List<Arg<?>> argsList = new ArrayList<>(parameters.size());
+        for (Entry<String, String> parameter : parameters.entrySet()) {
+            argsList.add(UnsafeArg.of(parameter.getKey(), parameter.getValue()));
+        }
+        return Collections.unmodifiableList(argsList);
```
dansanduleac: Since the `error.parameters()` are all unsafe, I'd just copy that to a new ImmutableMap and log the map itself as an `UnsafeArg`.

Then I'd just expose errorCode, errorName, errorInstanceId in here as SafeArgs

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
dansanduleac: It's fine for the RuntimeException message (passed to super) to be interpolated, but the logMessage shouldn't be.

pkoenig10: This doesn't have things interpolated it in, it's a `noArgsMessage`.

pkoenig10:
We explicitly decided not to inject args into `ServiceException` in https://github.com/palantir/conjure-java-runtime-api/pull/133 due to legibility concerns. There's also the potential for args conflicts (what if I create an arg named `errorCode`?).

I don't have strong opinions on the particular decision, but I think it's important that the behavior of `ServiceException` and `RemoteException` is consistent. So if we want to change this behavior, let's do it in a follow up and apply it to both `ServiceException` and `RemoteException`.

pkoenig10:
Moved to #337

PR #289, opened 2019-08-22
[improvement] Add 401 Unauthorized ErrorType

amrav:
Palantir's authentication service uses 401 in the expected way, but currently clients have to manually propagate these. This PR allows us to add default 401 propagation in `RemoteExceptionMapper`, similar to how we propagate 403s.


carterkozak:
Would you mind starting with the conjure specification? https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#34-conjure-errors

pkoenig10:
This should remain open

pkoenig10:
Moved to #356.

PR #348, opened 2019-10-10
fix #347: ServiceException retains causal `errorInstanceId`

### LINKED ISSUE #347: ServiceExceptions should retain errorInstanceId from causes
## What happened?

Applications often catch and rethrow exceptions, for example, a service may catch RemoteException, and throw a more specific ServiceException with a better ErrorType using the RemoteException as a cause. Today this generates a second `errorInstanceId` despite being a continuation of the same error. Debugging the error by looking up the `errorInstanceId` becomes an O(n) operation rather than O(1) where a single key provides a veiw of the failure across systems.
The current behavior is perplexing because allowing RemoteExceptions to be reach conjure  exception mappers _does_ result in the desired behavior, however in some cases (think failures from an executor) rethrowing the original exception directly makes it harder to trace execution, creating a new exception on the thread calling `Future::get` allows us to follow the failure to an endpoint.

## What did you want to happen?

ServiceException should traverse causes for an `errorInstanceId` before generating a new identifier.

carterkozak:
## Before this PR
See #347

## After this PR
==COMMIT_MSG==
fix #347: ServiceException retains causal `errorInstanceId`
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-348.v2.yml
+++ b/changelog/@unreleased/pr-348.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: 'fix #347: ServiceException retains causal `errorInstanceId`'
+  links:
+  - https://github.com/palantir/conjure-java-runtime-api/pull/348
--- a/changelog/@unreleased/pr-348.v2.yml
+++ b/changelog/@unreleased/pr-348.v2.yml
@@ -1,5 +1,8 @@
 type: improvement
 improvement:
-  description: 'fix #347: ServiceException retains causal `errorInstanceId`'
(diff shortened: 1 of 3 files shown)


changelog/@unreleased/pr-348.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: 'fix #347: ServiceException retains causal `errorInstanceId`'
```
ferozco: could we add a more detailed description of the change?

ferozco:
👍

changelog/@unreleased/pr-348.v2.yml
carterkozak: Sure thing

PR #353, opened 2019-10-14
Fix #226: RemoteException messages are safe loggable

### LINKED ISSUE #226: Make RemoteException SafeLoggable

carterkozak:
Previously RemoteException data was only recorded (safely) if the
exception was propagated to a conjure exception mapper. Many
use cases (exceptions from an executor, for example) do not result
in this behavior. Catching an exception and logging it should allow
us to look up the root cause by ID.

## After this PR
==COMMIT_MSG==
RemoteException messages are safe loggable
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -16,10 +16,15 @@
 
 package com.palantir.conjure.java.api.errors;
 
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeLoggable;
+import java.util.Collections;
+import java.util.List;
+
 /**
  * An exception thrown by an RPC client to indicate remote/server-side failure.
  */
-public final class RemoteException extends RuntimeException {
+public final class RemoteException extends RuntimeException implements SafeLoggable {
     private static final long serialVersionUID = 1L;
 
     private final SerializableError error;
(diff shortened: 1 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -46,4 +51,16 @@ public RemoteException(SerializableError error, int status) {
...
+    @Override
+    public String getLogMessage() {
+        return getMessage();
+    }
+
+    @Override
+    public List<Arg<?>> getArgs() {
+        // RemoteException explicitly does not support arguments because they have already been recorded
+        // on the service which produced the causal SerializableError.
+        return Collections.emptyList();
```
ferozco: I think it could be helpful for debugging if you expose the remote exceptions args since it gives you all the information in a single log line instead of having to jump between services

carterkozak: I don't have a strong opinion. #337 and others like it (if memory serves) have been stalled/rejected for ages due to contention around parameters, so I'd prefer to get the unobjectionable piece through separately.

carterkozak: If supported, I think all the parameters would need to be marked unsafe, which is not terribly helpful in most cases.

ferozco: Lets move forward with this, we can always add more information later

ferozco:
👍

PR #356, opened 2019-10-14
Add 401 Unauthorized ErrorType

pkoenig10:
## Before this PR

Can't throw errors with HTTP status code 401.

See https://github.com/palantir/conjure/pull/367 for context.

## After this PR

==COMMIT_MSG==
This change enables generating errors using the `UNAUTHORIZED` error type, which maps to HTTP status code 401.
==COMMIT_MSG==

## Potential complications

We need to make sure that existing clients can handle `RemoteException` with status code 401.
Most of our clients are java, and so at the very minimum we should check that the current conjure-java-runtime clients would [parse](https://github.com/palantir/conjure-java-runtime/blob/950e91d168fe86573604b962c4bcdddb251dde2e/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemoteExceptionResponseHandler.java#L61-L62) and [forward](https://github.com/palantir/conjure-java-runtime/blob/950e91d168fe86573604b962c4bcdddb251dde2e/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java#L54) it correctly, which it seems like they do (there's no special validation on either `SerializableError.errorCode()` or `RemoteException.getStatus()`).

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/ErrorType.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/ErrorType.java
@@ -36,6 +36,7 @@ public abstract class ErrorType {
             Pattern.compile(String.format("%s:%s", UPPER_CAMEL_CASE, UPPER_CAMEL_CASE));
 
     public enum Code {
+        UNAUTHORIZED(401),
         PERMISSION_DENIED(403),
         INVALID_ARGUMENT(400),
         NOT_FOUND(404),
@@ -54,6 +55,8 @@ public abstract class ErrorType {
         }
     }
 
+    public static final ErrorType UNAUTHORIZED =
+            createInternal(Code.UNAUTHORIZED, "Default:Unauthorized");
     public static final ErrorType PERMISSION_DENIED =
             createInternal(Code.PERMISSION_DENIED, "Default:PermissionDenied");
     public static final ErrorType INVALID_ARGUMENT =
--- a/changelog/@unreleased/pr-356.v2.yml
+++ b/changelog/@unreleased/pr-356.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add an `UNAUTHORIZED` error type
+  links:
+  - https://github.com/palantir/conjure-java-runtime-api/pull/356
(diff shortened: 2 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/ErrorType.java
```diff
@@ -36,6 +36,7 @@
             Pattern.compile(String.format("%s:%s", UPPER_CAMEL_CASE, UPPER_CAMEL_CASE));
 
     public enum Code {
+        UNAUTHORIZED(401),
```
carterkozak: I think there's ongoing discussion about the conjure IR change, but rolling this out is sufficiently decoupled from supporting `UNAUTHORIZED` in conjure-defined errors imo.

carterkozak: Others may feel differently, thoughts @ferozco @iamdanfox @dansanduleac?

gatesn: What's the use-case here? Usually a 401 is accompanied by a [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) header, how would we set that using a ServiceException?

Makes me think most use-cases should just handle the 401 case themselves.

carterkozak: 401: I don't know who you are
403: I know who you are and you aren't allowed in

401 isn't recoverable without a login, 403 means a particular operation was not allowed.

gatesn: Yeah, that's why the WWW-Authenticate header is required. It tells the client _how_ they should login: basic auth, bearer auth, etc. 

I wanted to know specifically what the use-case was, e.g. so the server can throw 401s? Or so a client can handle them? And in particular, how the server/client is planning to handle it _without_ access to that header.

pkoenig10: This has already been litigated in https://github.com/palantir/conjure/pull/367 and in internal discussions.

dansanduleac: Note that I have reverted that conjure PR, because it was an IR change and we can't release that yet until we've agreed on a process for how to make IR changes.

However, the addition of 401 itself makes sense, and I think we can still make this change on its own in conjure-java-runtime-api as long as the current conjure-java-runtime clients would forward it correctly, which it seems like [they do](https://github.com/palantir/conjure-java-runtime/blob/950e91d168fe86573604b962c4bcdddb251dde2e/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java#L54).

PR #368, opened 2019-10-23
Redact password and token-related fields

wwong:
Remove sensitive fields from toString output to prevent credential
leakage into logs and other similar places.

## Before this PR

Conjure consumers that log config classes might inadvertently leak passwords or tokens to disk

## After this PR
==COMMIT_MSG==
Password and token-related fields are now annotated as redacted, which should prevent the field values from appearing in toString output.
==COMMIT_MSG==

## Possible downsides?

Consumers that expect passwords and tokens to appear in toString output will now need to call the respective getters explicitly.

### DIFF
--- a/changelog/@unreleased/pr-368.v2.yml
+++ b/changelog/@unreleased/pr-368.v2.yml
@@ -1,5 +1,5 @@
-type: improvement
-improvement:
+type: fix
+fix:
   description: Password and token-related fields are now annotated as redacted, which
     should prevent the field values from appearing in toString output.
   links:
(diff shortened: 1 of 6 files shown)


changelog/@unreleased/pr-368.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
```
carterkozak: mind updating this to `fix`?

carterkozak:
Thanks for fixing this @wwong! Happy to merge+release once the changelog is updated.

wwong:
Looks like you beat me to it! Thanks for the assist :)

PR #369, opened 2019-10-24
Change ImmutablesStyle's retention policy

JiahuiJiang:
## Before this PR

Any class that's importing an immutable class with @ImmutablesStyle in the package, or directly with `@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)`, but doesn't add 'org.immutables.value.Value' as compile dependency will get a warning during compilation

## After this PR

==COMMIT_MSG==
Change immutablesStyle's retention policy to be SOURCE instead of CLASS, and make SerializableError use it. This avoids compilation warnings for consumers that don't depend on immutables.
==COMMIT_MSG==


errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
```diff
@@ -36,7 +36,8 @@
 @JsonDeserialize(builder = SerializableError.Builder.class)
 @JsonSerialize(as = ImmutableSerializableError.class)
 @Value.Immutable
-@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE, jdkOnly = true)
+@ImmutablesStyle
+@Value.Style(overshadowImplementation = false)
```
dansanduleac: this is to preserve binary compat, otherwise revapi helpfully complains :)

dansanduleac:
👍

errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
iamdanfox: do you know only this class was flagged for this and not all the others?

PR #370, opened 2019-10-25
Add utility to propagate RemoteExceptions

rshkv:
## Before this PR
In a request `A->B->C`, it's unreasonably difficult for B to propagate back a RemoteException it received from C. The common use case here is that C returns a 400 error that A could action, but B converts it to a `Default:Internal` error.

That's fine for the default. But if B identifies an error worth forwarding it has two cumbersome options:
* wrap RemoteExceptions in a different type and handle in a new ExceptionMapper
* re-throw RemoteExceptions as ServiceExceptions with a new error type and convert the SerializableError's parameters to UnsafeArgs

## After this PR
==COMMIT_MSG==
Adds utility to propagate original RemoteExceptions back to clients.
==COMMIT_MSG==
```java
try {
  remote.method()
} catch (RemoteException e) {
  ConjureThrowables.propagateIfErrorTypeEquals(e, RemoteErrors.ERROR);
  throw e;
}
```

### DIFF
--- a/changelog/@unreleased/pr-370.v2.yml
+++ b/changelog/@unreleased/pr-370.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+fix:
+  description: Adds utility to propagate original RemoteExceptions back to clients.
+  links:
+  - https://github.com/palantir/conjure-java-runtime-api/pull/370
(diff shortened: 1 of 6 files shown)


rshkv:
Just noted #337. I think merging #337 to have `getArgs()` on RemoteExceptions would make propagating errors much less painful:
```diff
try {
  service.method();
catch (RemoteException e) {
- Arg[] args = // 2-3 lines of turning parameters to args
- throw new ServiceException(ETYPE, e, args);
+ throw new ServiceException(ETYPE, e, e.getArgs().toArray(new Arg[0]);
}
```

PR #388, opened 2019-11-14
Log UnknownRemoteException body

mcopes73:

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
@@ -19,6 +19,7 @@ package com.palantir.conjure.java.api.errors;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.UnsafeArg;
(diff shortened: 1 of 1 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
```diff
@@ -54,7 +54,9 @@ public String getLogMessage() {
 
     @Override
     public List<Arg<?>> getArgs() {
-        return Collections.singletonList(SafeArg.of("status", getStatus()));
+        return Collections.singletonList(
+            SafeArg.of("status", getStatus()),
+            UnsafeArg.of("body", getBody()));
```
carterkozak: This is potentially very large, I'd actually prefer if we deprecated and removed the body field. In many cases the body cannot reasonably be expressed as a string.

gatesn: We used to do this before moving to UnknownRemoteExceptions here: https://github.com/palantir/conjure-java-runtime/blob/d29e227a2b553890e1b6b6cb8644d60ffb2e3cbf/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/IoExceptionResponseHandler.java#L49

PR #426, opened 2020-01-14
Improvement: Assume default type is PEM if trustStorePath/keyStorePath doesn't have jks extension

robert3005:
## Before this PR
We assume keystore and truststore type is jks regardlesss of what file it is

## After this PR
==COMMIT_MSG==
Try to be smarter about keystore and truststore type and only have them be jks if files end with jks extension
==COMMIT_MSG==

## Possible downsides?
Someone was relying on current behaviour for jks files without .jks extension

### DIFF
--- a/ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
+++ b/ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
@@ -35,8 +35,6 @@ public abstract class SslConfiguration {
         PUPPET
     }
 
-    private static final StoreType DEFAULT_STORE_TYPE = StoreType.JKS;
-
     @JsonAlias("trust-store-path")
     public abstract Path trustStorePath();
 
@@ -44,7 +42,11 @@ public abstract class SslConfiguration {
     @Value.Default
     @JsonAlias("trust-store-type")
     public StoreType trustStoreType() {
-        return DEFAULT_STORE_TYPE;
+        if (trustStorePath().getFileName().toString().endsWith("jks")) {
+            return StoreType.JKS;
+        } else {
+            return StoreType.PEM;
+        }
     }
 
     @JsonAlias("key-store-path")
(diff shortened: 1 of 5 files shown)


ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
```diff
@@ -35,16 +35,18 @@
...
-
     @JsonAlias("trust-store-path")
     public abstract Path trustStorePath();
 
     @SuppressWarnings("checkstyle:designforextension")
     @Value.Default
     @JsonAlias("trust-store-type")
     public StoreType trustStoreType() {
-        return DEFAULT_STORE_TYPE;
+        if (trustStorePath().getFileName().toString().endsWith("jks")) {
```
CRogers: For compatibility, might it make more sense to do the opposite? If it ends in `.pem` then it's a PEM otherwise it's a JKS? Then we also enter the PEM world forcing everyone to use a single `.pem` extension rather than whatever they want.

robert3005: annoyingly the opposite is multiple extensions where people use `pem`, `cer`, `key` but maybe that's small enough list?

robert3005:
@CRogers inverted the check to do what you suggested

ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
```diff
@@ -58,7 +63,13 @@ public StoreType trustStoreType() {
     @Value.Default
     @JsonAlias("key-store-type")
     public StoreType keyStoreType() {
-        return DEFAULT_STORE_TYPE;
+        if (keyStorePath().map(keyStore -> PEM_EXTENSIONS.stream()
+                .anyMatch(ext -> keyStore.getFileName().toString().endsWith(ext)))
```
CRogers: I'd just extract this `stream()...anyMatch()` logic to a method

CRogers:
:+1: given comment

ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
robert3005: Extracted it out

PR #427, opened 2020-01-14
Improvement: Only require keystore password for jks stores

robert3005:
## Before this PR
We required password for any kind of keystore

## After this PR
==COMMIT_MSG==
Don't require passwords for non jks keystores
==COMMIT_MSG==

## Possible downsides?
Code that uses those configurations might need to be altered but it doesn't break existing setups just relaxes constraints for future use

### DIFF
--- a/ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
+++ b/ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
@@ -67,9 +67,9 @@ public abstract class SslConfiguration {
 
     @Value.Check
     protected final void check() {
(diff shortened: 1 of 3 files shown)


ssl-config/src/main/java/com/palantir/conjure/java/api/config/ssl/SslConfiguration.java
```diff
@@ -67,9 +67,9 @@ public StoreType keyStoreType() {
 
     @Value.Check
     protected final void check() {
-        if (keyStorePath().isPresent() != keyStorePassword().isPresent()) {
+        if (keyStorePath().isPresent() && keyStoreType().equals(StoreType.JKS) && !keyStorePassword().isPresent()) {
```
carterkozak: bug: this allows a keystore password to be provided without a keystore.

PR #443, opened 2020-01-28
Add a new `enable-http2` flag to the service configuration 

carterkozak:
## Before this PR
No way to configure http/2 support in configuration

## After this PR
==COMMIT_MSG==
Add a new `enable-http2` flag to the service configuration 
==COMMIT_MSG==

## Possible downsides?
larger api surface.


carterkozak:
http/2 requires http/1 support in case negotiation fails, I don't think we want to add support for prior-knowledge http2.

I think the optional boolean provides the minimal complexity

carterkozak:
I'm going to hold off merging+releasing this until I've wired everything up in conjure-java-runtime to avoid confusion. Once we're happy with both sides we can release them in close succession.

Issue #160, opened 2018-11-26
ProxyConfiguration option for using JVM default proxy selector


gatesn:
I would like a way to configure a service to use the JVM's default proxy selector. Currently if `ProxyConfiguration#type = HTTP`, it bypasses the system settings. It looks like `ProxyConfiguration#type = MESH` uses the default selector, but it seems wrong to configure that.

What are your thoughts on either:

1. Introduce a new `type = DEFAULT`
1. Default `hostAndPort` to `System.getenv("http_proxy")` if not set for `type = HTTP`

gatesn:
Ping on this. Would like to know which option, if any, I should go ahead with?

LorenzoMartini:
Solved in #535

PR #535, opened 2020-04-06
Add ProxySelection type FROM_ENVIRONMENT

LorenzoMartini:
Addresses #160.
Add new proxy selection type "FROM_ENVIRONMENT" that will be used in `conjure-java-runtime` to instantiate proxy selection based on environment variable `https_proxy`.

https://github.com/palantir/conjure-java-runtime/pull/1566 for the change in `c-j-r`.

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
@@ -38,6 +38,9 @@ public abstract class ProxyConfiguration {
 
     public enum Type {
 
+        /** Use default JVM's proxy selector. */
+        DEFAULT,
+
         /** Use a direct connection. This option will bypass any JVM-level configured proxy settings. */
         DIRECT,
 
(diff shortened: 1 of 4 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
```diff
@@ -41,6 +41,9 @@
         /** Use a direct connection. This option will bypass any JVM-level configured proxy settings. */
         DIRECT,
 
+        /** Use default JVM's proxy selector. */
```
gatesn, line 44: This comment is accurate

service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
```diff
@@ -93,6 +96,11 @@ protected final void check() {
                         !hostAndPort().isPresent() && !credentials().isPresent(),
                         "Neither credential nor host-and-port may be configured for DIRECT proxies");
                 break;
+            case FROM_ENVIRONMENT:
+                Preconditions.checkArgument(
+                        !hostAndPort().isPresent() && !credentials().isPresent(),
+                        "Neither credential nor host-and-port may be configured for FROM_ENVIRONMENT proxies");
```
LorenzoMartini, line 102: Not sure about this, do we need credentials in some cases for this?

iamdanfox: I'd leave credentials out tbh, as in we might want to support authenticated proxies at some point. https://stackoverflow.com/a/1626616/1941552

validating no hostAndPort still makes sense though, as users specifying these _definitely_ doesn't make sense in the FROM_ENVIRONMENT case.

LorenzoMartini: 👍

PR #540, opened 2020-04-15
Add Helper for PartialServiceConfiguration

LorenzoMartini:
## Before this PR

There is no way of creating a `PartialServiceConfiguration` object from a `ServiceConfiguration` one even though they share all the fields.
The fields all have the same name except `proxy` and some have slightly different types and that makes it problematic to just switch between the two types without having to match all the fields one by one.

## After this PR
==COMMIT_MSG==
Add Helper class to create PartialServiceConfiguration from ServiceConfiguration.
==COMMIT_MSG==


robert3005:
The fact you're going ServiceConfiguration -> PartialServiceConfiguration suggests there's something very weird with what you're trying to do. 👎

PR #563, opened 2020-06-01
unsafe log unknown remote exception body

ferozco:
## Before this PR
We captured the body of an unknown remote exception and promptly ignored it.

## After this PR
==COMMIT_MSG==
add the body of the UnknownRemoteException as a unsafe parameter 
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
@@ -19,6 +19,8 @@ package com.palantir.conjure.java.api.errors;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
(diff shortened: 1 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
```diff
@@ -52,6 +54,9 @@ public String getLogMessage() {
 
     @Override
     public List<Arg<?>> getArgs() {
-        return Collections.singletonList(SafeArg.of("status", getStatus()));
+        List<Arg<?>> args = new ArrayList<>(2);
+        args.add(SafeArg.of("status", getStatus()));
+        args.add(UnsafeArg.of("body", getBody()));
+        return Collections.unmodifiableList(args);
```
carterkozak, line 60: nit: maybe `Collections.unmodifiableList(Arrays.asList(SafeArg.of("status", getStatus()),UnsafeArg.of("body", getBody())))`

PR #590, opened 2020-07-15
HasExternalSource

j-baker:
Add a simple interface that can be used to mark exceptions as being caused by an external problem (e.g. a third service failing). This enables us to internally monitor each service for exceptions that _aren't_ caused by an external problem (higher signal). This means that a library like AtlasDB can trivially make its AtlasDbDependencyException (thrown when e.g. Cassandra is broken) be picked up as an external problem.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -22,7 +22,7 @@ import java.util.Collections;
 import java.util.List;
 
 /** An exception thrown by an RPC client to indicate remote/server-side failure. */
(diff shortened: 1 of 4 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -22,7 +22,7 @@
 import java.util.List;
 
 /** An exception thrown by an RPC client to indicate remote/server-side failure. */
-public final class RemoteException extends RuntimeException implements SafeLoggable {
+public final class RemoteException extends RuntimeException implements SafeLoggable, HasExternalSource {
```
carterkozak, line 25: Does this mean constructing bad requests which result in 4xx responses becomes somebody else's problem?

PR #591, opened 2020-07-15
Support QosException with cause

pkoenig10:
## Before this PR
Users can't create a `QosException` with a cause. This means we lose the cause information is lost if a caught exception is re-thrown as a `QosException`.

## After this PR
Users can create a `QosException` with a cause.

### DIFF
--- a/changelog/@unreleased/pr-591.v2.yml
+++ b/changelog/@unreleased/pr-591.v2.yml
@@ -1,0 +1,5 @@
+type: feature
+feature:
+  description: '`QosException` can now be created with a cause.'
+  links:
+  - https://github.com/palantir/conjure-java-runtime-api/pull/591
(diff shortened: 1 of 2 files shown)


markelliot:
What would a consumer do with this information? If it’s for service debugging, I think we’d probably prefer logging the relevant cause information at the catch site rather than depending on the outbound server logging recording cause information.

pkoenig10:
Exception causes are not included in `SerializableError` and never reach consumers. This is purely for logging the exception.

> I think we’d probably prefer logging the relevant cause information at the catch site rather than depending on the outbound server logging recording cause information.

We support providing a cause for the exceptions in https://github.com/palantir/safe-logging. It's seems weird that the rationale here would not be the same.

Sure, explicitly logging at the catch site is a solution. But it just seems unnecessarily cumbersome.

Do you have a specific concern with exposing this functionality? Is there some behavioral or API design risk?

carterkozak:
I also prefer retaining causes and providing full context at the top level (service exception handler in this case) unless handling an exception is expected to take a great deal of time. That way we don't have to worry about updating N projects log statements if we decide qos exceptions should be logged at info instead of debug. catch + log + throw is an antipattern because it's harder to piece together full context at a glance. Take the active project to aggregate internal error traces for example, the resulting stack traces are more actionable if they provide the full cause chain.

markelliot:
Ok.

PR #604, opened 2020-08-27
improvement: move errorInstanceId from exception message into exception args

ferozco:
## Before this PR
We were attempting to perform some internal analysis on unique stacktraces but were running into issues since we grouped stacktraces textually and the error instance id appeared as part of the message of each remote exception

## After this PR

==COMMIT_MSG==
move `RemoteException`s errorInstanceId from exception message into exception args 
==COMMIT_MSG==

## Possible downsides?
It will be harder to debug issue if logging is not set up (i.e. args are logged) correctly since the message will not contain the error id

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.api.errors;
 
 import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -41,14 +43,12 @@ public int getStatus() {
     public RemoteException(SerializableError error, int status) {
         super(
                 error.errorCode().equals(error.errorName())
-                        ? String.format(
-                                "RemoteException: %s with instance ID %s", error.errorCode(), error.errorInstanceId())
-                        : String.format(
-                                "RemoteException: %s (%s) with instance ID %s",
-                                error.errorCode(), error.errorName(), error.errorInstanceId()));
+                        ? String.format("RemoteException: %s", error.errorCode())
+                        : String.format("RemoteException: %s (%s)", error.errorCode(), error.errorName()));
```
carterkozak, line 47: Let's avoid modifying the result of `getMessage` and only update `getLogMessage`

carterkozak: There are a few consumers that don't use safe-logging aware layouts, I don't want to make debugging more difficult for them.

ferozco: good catch! this should give us the best of both worlds

PR #608, opened 2020-09-10
ServiceException no longer throws NPE when null args are provided

carterkozak:
## Before this PR
The render method threw a NPE when an argument value was null. This exception destroys our ability to capture the root cause -- we must either set the original as a cause/suppressed-cause or provide as much data as possible.

It's clearly not ideal to have null arg values, however it's even worse if we're unable to diagnose the original problem based on logging, so we mustn't be opinionated here.

## After this PR

==COMMIT_MSG==
ServiceException no longer throws NPE when null args are provided
==COMMIT_MSG==

See https://github.com/palantir/safe-logging/pull/492

## Possible downsides?
none.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
@@ -95,27 +95,37 @@ public final class ServiceException extends RuntimeException implements SafeLogg
     }
 
     private static <T> List<T> copyToUnmodifiableList(T[] elements) {
+        if (elements == null || elements.length == 0) {
+            return Collections.emptyList();
+        }
         List<T> list = new ArrayList<>(elements.length);
-        Collections.addAll(list, elements);
+        for (T item : elements) {
+            if (item != null) {
+                list.add(item);
+            }
+        }
         return Collections.unmodifiableList(list);
     }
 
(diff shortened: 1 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java
```diff
@@ -95,27 +95,37 @@ public String getLogMessage() {
...
-            if (i > 0) {
-                builder.append(", ");
+        for (Arg<?> arg : args) {
+            if (arg != null) {
+                if (first) {
+                    first = false;
+                } else {
+                    builder.append(", ");
+                }
+                builder.append(arg.getName()).append("=").append(arg.getValue());
```
fawind, line 127: Do we want to keep the information that an arg was null in the message? E.g. rendering something like `{arg1=null}`? That way you could see that something went wrong passing through the args when inspecting the logs.

carterkozak: I don't think that tells us anything, and we would have to use a funky argument name to prevent duplication  with user-provided arg names.

fawind: Makes sense!

fawind (approved):

carterkozak:
:robot::popcorn:

PR #612, opened 2020-10-08
All QosExceptions are SafeLoggable

iamdanfox:
## Before this PR

I was just looking at an internal stacktrace and saw the message was redacted, which made me wonder what it was:

```
Caused by: com.palantir.conjure.java.api.errors.QosException$Throttle: {throwable1_message}
	at com.palantir.conjure.java.api.errors.QosException.throttle(QosException.java:60)
	at java.util.Optional.orElseGet(Optional.java:369)
	at com.palantir.conjure.java.dialogue.serde.ErrorDecoder.decode(ErrorDecoder.java:88)
```

Turns out there was nothing interesting in this, but it made me curious enough that I wasted 2 minutes opening up the QosException.java class.

## After this PR
==COMMIT_MSG==
All QosExceptions are now SafeLoggable
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/QosException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/QosException.java
@@ -19,9 +19,10 @@ package com.palantir.conjure.java.api.errors;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.UnsafeArg;
 import java.net.URL;
 import java.time.Duration;
-import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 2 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/QosException.java
```diff
@@ -167,25 +178,35 @@ public String getLogMessage() {
 
         @Override
         public List<Arg<?>> getArgs() {
-            List<Arg<?>> args = new ArrayList<>();
-            args.add(SafeArg.of("redirectTo", redirectTo));
```
iamdanfox, line 171: I've flipped this to an UnsafeArg too.

I'm pretty sure this codepath is unused at the moment because none of our services actually send the redirect-to header, but figured we might as well fix it at the same time!

carterkozak: :+1:

carterkozak (approved):

PR #633, opened 2021-02-23
RemoteException.getMessage includes SerializableError parameters

carterkozak:
In connected environments this makes no difference because
RemoteException.getMessage is never called, only getLogMessage.
In such cases the error instance identifier is sufficient to
search for the original failure information, however service
authors without safe-logging infrastructure are unlikely to have
access to remote instance logging and should be presented with
the failure parameters in local logging.

The message is computed lazily, very slightly reducing overhead
in environments that don't use the unsafe message, while providing
additional actionable signal to others.

## After this PR

==COMMIT_MSG==
RemoteException.getMessage (unsafe message) includes SerializableError parameters. RemoteException.getLogMessage (safe) is not impacted.
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -26,11 +26,13 @@ import java.util.List;
 public final class RemoteException extends RuntimeException implements SafeLoggable {
     private static final long serialVersionUID = 1L;
 
-    private final String message;
     private final String stableMessage;
     private final SerializableError error;
     private final int status;
     private final List<Arg<?>> args;
+    // Lazily evaluated based on the stableMessage, errorInstanceId, and args.
+    @SuppressWarnings("MutableException")
+    private String unsafeMessage;
 
     /** Returns the error thrown by a remote process which caused an RPC call to fail. */
     public SerializableError getError() {
@@ -44,9 +46,8 @@ public final class RemoteException extends RuntimeException implements SafeLogga
 
     public RemoteException(SerializableError error, int status) {
         this.stableMessage = error.errorCode().equals(error.errorName())
-                ? String.format("RemoteException: %s", error.errorCode())
-                : String.format("RemoteException: %s (%s)", error.errorCode(), error.errorName());
-        this.message = this.stableMessage + " with instance ID " + error.errorInstanceId();
+                ? "RemoteException: " + error.errorCode()
+                : "RemoteException: " + error.errorCode() + " (" + error.errorName() + ")";
         this.error = error;
         this.status = status;
         this.args = Collections.singletonList(SafeArg.of("errorInstanceId", error.errorInstanceId()));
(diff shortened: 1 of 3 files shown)


pkoenig10:
Just want to flag that we previously rejected this change in https://github.com/palantir/conjure-java-runtime-api/pull/337. I'm not opposed to making this change (I think I'd actually prefer it), I just want to ensure that we're now willing to accept with the tradeoffs here.

carterkozak:
Yep yep, Felix reported a case where an issue was difficult to debug due to lack of insight into the server-side of the problem, recorded in a non-safe-logging environment. I think that’s the upside we were missing last time.

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -44,17 +46,40 @@ public int getStatus() {
...
+                .append(stableMessage)
+                .append(" with instance ID ")
+                .append(error.errorInstanceId());
+        if (!error.parameters().isEmpty()) {
+            builder.append(": {");
+            error.parameters()
+                    .forEach((name, unsafeValue) ->
+                            builder.append(name).append('=').append(unsafeValue).append(", "));
+            // remove the trailing space
+            builder.setLength(builder.length() - 1);
```
jkozlowski, line 78: Joiner? Does that allocate?

carterkozak: In this case it's unlikely that the last character resulted in a buffer expansion, so reducing the length by 1 is a non-volatile write to the StringBuilder.count integer.

Joiner would work nicely if this didn't involve relatively complex prefix parameters -- we'd end up creating a bunch of intermediate strings.

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -44,17 +46,40 @@ public int getStatus() {
...
         this.error = error;
         this.status = status;
         this.args = Collections.singletonList(SafeArg.of("errorInstanceId", error.errorInstanceId()));
     }
 
     @Override
     public String getMessage() {
-        return message;
+        // This field is not used in most environments so the cost of computation may be avoided.
+        String messageValue = unsafeMessage;
```
fawind, line 59: Curious, why do we need the `messageValue` helper variable instead of doing something like:
```java
if (unsafeMessage == null) {
    unsafeMessage = renderUnsafeMessage();
}
return unsafeMessage;
```

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -44,17 +46,40 @@ public int getStatus() {
...
-        this.message = this.stableMessage + " with instance ID " + error.errorInstanceId();
+                ? "RemoteException: " + error.errorCode()
+                : "RemoteException: " + error.errorCode() + " (" + error.errorName() + ")";
         this.error = error;
         this.status = status;
         this.args = Collections.singletonList(SafeArg.of("errorInstanceId", error.errorInstanceId()));
     }
 
     @Override
     public String getMessage() {
```
fawind, line 57: Should we also add the new safety annotations to this class?

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
carterkozak: Optimization, not likely a measurable difference, but we’re reading the field value once instead of twice. I find it cleaner to assert on the same value that’s returned as well (avoids a class of race condition bugs)

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
carterkozak: Wouldn’t hurt, although we already consider throwable.getMessage unsafe across the board

fawind (approved):
Looks good, especially given that even previously this was positively received, just lacking an incentive.

PR #655, opened 2021-04-28
RemoteException emits more safeargs

iamdanfox:
## Before this PR

Context in #dev-rpc - BLUF is that logs in aries are hard to filter when there are many RemoteExceptions, but they have different causes.

## After this PR
==COMMIT_MSG==
RemoteExceptions now emit SafeArgs for `errorInstanceId`, `errorCode` and `errorName` (previously it was just `errorInstanceId`)
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
@@ -19,12 +19,16 @@ package com.palantir.conjure.java.api.errors;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -49,7 +53,10 @@ public RemoteException(SerializableError error, int status) {
         this.message = this.stableMessage + " with instance ID " + error.errorInstanceId();
         this.error = error;
         this.status = status;
-        this.args = Collections.singletonList(SafeArg.of("errorInstanceId", error.errorInstanceId()));
+        this.args = Collections.unmodifiableList(Arrays.asList(
```
iamdanfox, line 56: `ImmutableList` is off the cards cos this project doens't seem to depend on guava, and we still have sourceCompat = 1.8 here so no `List.of` either.

carterkozak (approved):

iamdanfox:
Thank you!

PR #659, opened 2021-06-22
Clarify UnknownRemoteException message

pkoenig10:
## Before this PR
The `UnknownRemoteException` error message can be misleading. Dialogue uses `UnknownRemoteException` for things other than just failing to deserialize a `SerializableError`.

## After this PR
The `UnknownRemoteException` error message has been made a bit more generic to avoid being misleading.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
@@ -42,7 +42,7 @@ public final class UnknownRemoteException extends RuntimeException implements Sa
     }
 
     public UnknownRemoteException(int status, String body) {
-        super(String.format("UnknownRemoteException: %s", status));
+        super(String.format("Response status: %s", status));
         this.status = status;
         this.body = body;
     }
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
@@ -42,14 +42,14 @@ public final class UnknownRemoteException extends RuntimeException implements Sa
     }
 
     public UnknownRemoteException(int status, String body) {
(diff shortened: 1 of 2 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
```diff
@@ -42,14 +42,14 @@ public String getBody() {
     }
 
     public UnknownRemoteException(int status, String body) {
-        super(String.format("Error %s. (Failed to parse response body as SerializableError.)", status));
+        super(String.format("UnknownRemoteException: %s", status));
```
pkoenig10, line 45: Tried to match the format of `RemoteException` and `ServiceException`.

carterkozak:
> Dialogue uses UnknownRemoteException for things other than just failing to deserialize a SerializableError.

What are the other scenarios UnknownRemoteException is thrown?

pkoenig10:
This is the particular instance that prompted this PR.
https://github.com/palantir/dialogue/blob/33443f62c983c16eaacc8c68b551f878c35eeb1c/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java#L84

carterkozak:
Makes sense, that case certainly cannot be parsed into a SerializableError, but the current message is confusing and makes it seem like something has gone wrong in an unexpected way.  Lately we've seen proxies produce non-serializableerror non-2xx responses, and the wording makes folks thing there's a conjure client bug rather than a problem between services.

For the wording, I'm not sure it makes sense to duplicate the class name in the message, our stack traces already include class name. Perhaps something along the lines of `Response status <CODE>` to make it more obvious the number is a a response status?

pkoenig10:
> Perhaps something along the lines of `Response status <CODE>` to make it more obvious the number is a a response status?

Sure that seems reasonable. I was only trying to be consistent with the other exception messages which have the form `ClassName: <errorCode>`.

carterkozak (approved):

PR #696, opened 2021-09-28
add UserAgents::tryParse(RequestContext) mtod

blakehawkins:
## Before this PR
Small hoops need jumping through, i.e.

```
    private static UserAgent extractUserAgent(RequestContext requestContext) {
        return UserAgents.tryParse(requestContext.firstHeader(HttpHeaders.USER_AGENT).orElse(null));
    }
```

## After this PR

==COMMIT_MSG==
UserAgents::tryParse now has convenience overload with RequestContext as param
==COMMIT_MSG==

## Possible downsides?
- [ ] adds a dep on conjure-undertow-lib


service-config/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     api "com.palantir.safe-logging:preconditions"
     api "com.palantir.tokens:auth-tokens"
     implementation "com.palantir.safe-logging:logger"
+    implementation 'com.palantir.conjure.java:conjure-undertow-lib:6.5.0'
```
carterkozak, line 12: I don't think we want this module to depend on specific conjure implementations -- at the limit it would need overloads for every conjure implementation. Perhaps we could invert the dependency, or perhaps cleaner yet, implement the logic in a third library which depends on service-config and conjure-undertow-lib.

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -73,6 +75,10 @@ public static UserAgent parse(String userAgent) {
         return parseInternal(userAgent, false /* strict */);
     }
 
+    public static UserAgent tryParse(RequestContext requestContext) {
+        return tryParse(requestContext.firstHeader(HttpHeaders.USER_AGENT).orElse(null));
```
carterkozak, line 79: I don't think this is quite right, our typescript client cannot override the browser `User-Agent` so it uses a custom header (iirc `Fetch-User-Agent` or similar) which must be checked first.

carterkozak (changes requested):
I don't think this is the right place for this functionality to live

blakehawkins:
Thanks @carterkozak , let's close this as I don't think a bunch of dependency inversion is right for this small util

PR #705, opened 2021-10-29
Add FieldMissingException

viktornordling:
@CRogers, this has been discussed with @carterkozak. The goal is for clients to receive a more descriptive error than a `Default:InvalidArgument` when a json payload is sent which is missing fields.

We will first add this exception here, then optionally (if a feature flag is set) use the exception (in [this PR](https://github.com/palantir/conjure-java/pull/1522)) and finally we will add an exception mapper, in [this PR](https://github.com/palantir/conjure-java-runtime/pull/2142).

==COMMIT_MSG==
Add FieldMissingException
==COMMIT_MSG==


viktornordling:
@CRogers, this has been discussed with @carterkozak. The goal is for clients to receive a more descriptive error than a `Default:InvalidArgument` when a json payload is sent which is missing fields.

PR #710, opened 2021-11-03
Construct RemoteException from ServiceException

tdeitch:
## Before this PR

When testing how my code handles RemoteExceptions from other services, I often construct RemoteExceptions from other services' published Conjure ServiceExceptions. I always end up writing the same little utility class to serialize the ServiceException, get its error code, and use that to create a RemoteException.

## After this PR

==COMMIT_MSG==
Construct RemoteException from ServiceException
==COMMIT_MSG==

### DIFF
--- a/errors/src/test/java/com/palantir/conjure/java/api/errors/RemoteExceptionTest.java
+++ b/errors/src/test/java/com/palantir/conjure/java/api/errors/RemoteExceptionTest.java
@@ -47,6 +47,15 @@ public final class RemoteExceptionTest {
     }
 
     @Test
+    public void testFromServiceException() {
+        ServiceException serviceException =
+                new ServiceException(ErrorType.INTERNAL, SafeArg.of("safeArg", "safeValue"));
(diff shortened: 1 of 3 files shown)


errors/src/test/java/com/palantir/conjure/java/api/errors/RemoteExceptionTest.java
```diff
@@ -46,6 +46,15 @@ public void testJavaSerialization() {
         assertThat(actual).isEqualToComparingFieldByField(expected);
     }
 
+    @Test
+    public void testFromServiceException() {
+        ServiceException serviceException =
+                new ServiceException(ErrorType.INTERNAL, SafeArg.of("safeArg", "safeValue"));
+        RemoteException expected = new RemoteException(SerializableError.forException(serviceException), 500);
```
tdeitch, line 53: I'm cognizant that this is basically a repeat of the implementation so can change it to something that tests the serialization explicitly, but on the other hand I didn't want to re-test `SerializableError`.

PR #864, opened 2022-07-14
Optimize UserAgent version validation

schlosna:
## Before this PR
`com.palantir.conjure.java.api.config.service.UserAgents#isValidVersion` is frequently used to validate the user agent version string, and currently uses an expensive regex which does a bunch of allocations and expensive processing. This generates unnecessary garbage and CPU cycles for services handling high volumes of HTTP requests.

image
image

## After this PR
We can use a hand rolled parser borrowed and simplified from https://github.com/palantir/sls-version-java/pull/463 to validate `UserAgent` version strings up to 4 groups (commonly used by browsers).

==COMMIT_MSG==
Optimize UserAgent version validation
==COMMIT_MSG==

## Possible downsides?
Duplicates code locally from https://github.com/palantir/sls-version-java because browser version strings don't conform to SLS versioning.

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/VersionParser.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/VersionParser.java
@@ -24,11 +24,11 @@ import com.google.errorprone.annotations.Immutable;
  * browser user-agent version strings.
  * <p>
  * We're using parser combinator-style ideas here, where each parser function
- * {@link #number(String, int)} accepts an index into our source string and returns a two values:
+ * {@link #number(String, int)} accepts an index into our source string and returns two values:
  * <p>
  * - an updated index into the string representing how many characters were parsed
  * - a value that was actually parsed out of the string (if the parser was able to parse the string), otherwise a
- * clear signal that the parser failed (we use {@link Integer#MIN_VALUE} for this.
+ * clear signal that the parser failed (we use {@link Integer#MIN_VALUE} for this).
  * <p>
  * We bit-pack these two integer values into a single long using {@link #ok(int, int)} and {@link #fail(int)} functions
  * because primitive longs live on the stack and don't impact GC.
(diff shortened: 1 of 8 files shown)


service-config/src/test/java/com/palantir/conjure/java/api/config/service/VersionParserTest.java
```diff
@@ -0,0 +1,50 @@
...
+                Arguments.of("bad", -1, false),
+                Arguments.of("1", 1, true),
+                Arguments.of("1.2", 2, true),
+                Arguments.of("1.2.3", 3, true),
+                Arguments.of("1.2.3.4", 4, true),
+                Arguments.of("1.2.3-rc4", -1, true),
+                Arguments.of("1.2.3-4-gabc", -1, true),
+                Arguments.of("1.2.3.4-rc5-6-gabc", -1, true),
+                Arguments.of("1.2.3-4", -1, false),
+                Arguments.of("0-0-0", -1, false));
```
carterkozak, line 48: Might be helpful to add some test data with multi-digit numeric components e.g. `123.456.789` and `123.456.789-rc0`

service-config/src/main/java/com/palantir/conjure/java/api/config/service/VersionParser.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+@Immutable
+final class VersionParser {
+    private VersionParser() {}
+
+    /** Returns the count of version number groups parsed if a valid version string, or -1 otherwise. */
+    public static int countNumericDotGroups(String string) {
+        long state = 0;
+        for (int i = 1; getIndex(state) < string.length(); i++) {
+            state = number(string, getIndex(state));
```
carterkozak, line 44: We could probably dedup out repeated calls to `getIndex(state)` in places that `state` doesn't change, however it's a trivial operation and likely not worth deconstructing the loop

carterkozak (approved):

service-config/src/main/java/com/palantir/conjure/java/api/config/service/VersionParser.java
schlosna: I think I'm going to leave it as-is for now. Deduping the `getIndex` for loop condition & first use doesn't really change benchmark, so going to opt for simplicity and consistent use of `getIndex(state)`

service-config/src/test/java/com/palantir/conjure/java/api/config/service/VersionParserTest.java
schlosna: added a few more

PR #882, opened 2022-08-08
Add causes to assertion errors

tdeitch:
## Before this PR

Debugging assertions that fail due to a different exception being thrown doesn't give you enough information. For example, you might be expecting a service exception and see that a remote exception is thrown. What was the error name? There's no easy way to tell.

## After this PR

==COMMIT_MSG==
Add causes to assertion errors
==COMMIT_MSG==

## Possible downsides?

The main hypothetical question I had was: is this _actually_ a cause of the AssertionError, or am I just shoving an error where it's convenient? I would argue that this really is the cause of the AssertionError. If I'm expecting a ServiceException, but some RemoteException gets thrown instead, the RemoteException really did keep my code from reaching the ServiceException I was trying to trigger.

### DIFF
--- a/changelog/@unreleased/pr-882.v2.yml
+++ b/changelog/@unreleased/pr-882.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: Add causes to the AssertionErrors thrown by `assertThatServiceExceptionThrownBy`
+    and `assertThatRemoteExceptionThrownBy` if a throwable of the wrong class is raised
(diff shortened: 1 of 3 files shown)


carterkozak:
I agree with the idea, but I think we can produce behavior more consistent with other assertj checks using a slightly different tact. What do you think of #883?

tdeitch:
Oh nice, that works for me, closing this in favor of that.

PR #883, opened 2022-08-08
conjure assertions provide diagnostic info for unexpected exceptions

carterkozak:
This is based on the idea provided by #882, implemented to use
standard assertj types and provide feedback consistent with builtin
assertj checks.

==COMMIT_MSG==
conjure assertions provide diagnostic info for unexpected exceptions
==COMMIT_MSG==

### DIFF
--- a/test-utils/src/main/java/com/palantir/conjure/java/api/testing/ServiceExceptionAssert.java
+++ b/test-utils/src/main/java/com/palantir/conjure/java/api/testing/ServiceExceptionAssert.java
@@ -41,7 +41,7 @@ public class ServiceExceptionAssert extends AbstractThrowableAssert<ServiceExcep
     public final ServiceExceptionAssert hasArgs(Arg<?>... args) {
         isNotNull();
 
-        AssertableArgs actualArgs = new AssertableArgs(actual.getParameters());
+        AssertableArgs actualArgs = new AssertableArgs(actual.getArgs());
         AssertableArgs expectedArgs = new AssertableArgs(Arrays.asList(args));
 
         failIfNotEqual("Expected safe args to be %s, but found %s", expectedArgs.safeArgs, actualArgs.safeArgs);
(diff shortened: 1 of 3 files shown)


test-utils/src/main/java/com/palantir/conjure/java/api/testing/ServiceExceptionAssert.java
```diff
@@ -73,7 +73,7 @@ private void failIfNotEqual(String message, Object expected, Object actual) {
     public final ServiceExceptionAssert containsArgs(Arg<?>... args) {
         isNotNull();
 
-        AssertableArgs actualArgs = new AssertableArgs(actual.getParameters());
+        AssertableArgs actualArgs = new AssertableArgs(actual.getArgs());
```
carterkozak, line 76: Migrated away from deprecated methods

test-utils/src/test/java/com/palantir/conjure/java/api/testing/AssertionsTest.java
```diff
@@ -33,16 +33,16 @@ public void testAssertThatServiceExceptionThrownBy_failsIfNothingThrown() {
...
 
     @Test
     public void testAssertThatServiceExceptionThrownBy_failsIfWrongExceptionThrown() {
         assertThatThrownBy(() -> assertThatServiceExceptionThrownBy(() -> {
                     throw new RuntimeException();
                 }))
-                .hasMessage("Expecting code to throw a com.palantir.conjure.java.api.errors.ServiceException,"
-                        + " but caught a java.lang.RuntimeException.");
+                .hasMessageContaining(
+                        "com.palantir.conjure.java.api.errors.ServiceException", "java.lang.RuntimeException");
```
carterkozak, line 45: Generified the asserts since it's not generally helpful to test precise string matches.

tdeitch: I don't care about the precise string, but maybe to test that we are serializing the underlying error messages:

        assertThatThrownBy(() -> assertThatServiceExceptionThrownBy(() -> {
                    throw new RuntimeException("My helpful message");
                }))
                .hasMessageContaining(
                        "com.palantir.conjure.java.api.errors.ServiceException",
                        "java.lang.RuntimeException",
                        "My helpful message");

jkozlowski: +1

jkozlowski (approved):
I would support checking that the full message is serialized, as that's the point of this change, but I don't know the assertj implementation. If that's what it's basically supposed to do, then let's ship it, no need to re-test their implementation.

test-utils/src/test/java/com/palantir/conjure/java/api/testing/AssertionsTest.java
carterkozak: :+1:

PR #902, opened 2022-10-01
Add UserAgents.of(UserAgent, Iterable<Agent>)

schlosna:
## Before this PR
Creating a `UserAgent` with multiple additional informational agents required several immutable object creations. See https://github.com/palantir/dialogue/pull/1797 for example

## After this PR
==COMMIT_MSG==
Add UserAgents.of(UserAgent, Iterable<Agent>)
==COMMIT_MSG==

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.java.api.config.service;
 
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.List;
 import java.util.Optional;
 import org.immutables.value.Value;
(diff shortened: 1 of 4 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
```diff
@@ -86,13 +87,14 @@ interface Agent {
...
-                    "Illegal version format. This is a bug",
-                    SafeArg.of("version", version()));
+            if (!UserAgents.isValidName(name())) {
+                throw new SafeIllegalArgumentException("Illegal agent name format", SafeArg.of("name", name()));
+            }
+            if (!UserAgents.isValidVersion(version())) {
+                // Should never hit the following.
+                throw new SafeIllegalArgumentException(
+                        "Illegal version format. This is a bug", SafeArg.of("version", version()));
+            }
```
carterkozak, line 97: 👍

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -46,6 +47,16 @@ public final class UserAgents {
...
 
+    /**
+     * Returns a user agent with the given base agent combined with specified additional informational agents.
+     */
+    public static UserAgent of(UserAgent base, Iterable<Agent> additional) {
+        return ImmutableUserAgent.builder()
+                .from(base)
+                .addAllInformational(additional)
+                .build();
+    }
```
carterkozak, line 58: This is a bit odd given other factory methods live in the `UserAgent` class, however those don't take a base UserAgent.

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
carterkozak: We should probably update this to use immutables params and avoid the Agent.Builder altogether

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
schlosna: good call, not sure why I put it in `UserAgents` instead of with other factories in `UserAgent`. 

I'm not sure if using `@Value.Parameter` buys us a ton in terms of either few allocations or readability as we don't currently include Guava in the `service-config` module, so no cheap `Iterables.concat`:

```
    static UserAgent of(UserAgent base, Iterable<Agent> additional) {
        return ImmutableUserAgent.of(
                base.nodeId(),
                base.primary(),
                Stream.concat(base.informational().stream(), StreamSupport.stream(additional.spliterator(), false))
                        .collect(Collectors.toList()));
    }
```

carterkozak: You're right -- lgtm!

carterkozak:
👍

PR #920, opened 2022-11-21
Reduce noisy user agent logging to debug

pkoenig10:
We started seeing a ton of these logs. This log probably is not appropriate for `warn`.

Screen Shot 2022-11-21 at 2 34 16 PM


carterkozak (approved):
Thanks, @pkoenig10!

PR #925, opened 2022-12-08
Add a reason field to QosException

mpritham:
## Before this PR
Services throw `QosException`s to protect themselves from becoming overloaded. It's useful to know why a service responded with a `QosException` for better observability. However, the exceptions are currently logged at the `debug` level, which is disabled by default. It would be costly to log the exceptions at an `info` (or higher) level. 

This PR introduces new `QosException` factory methods which allow users to store a reason (an object of newly defined type `QosReason`) for throwing an exception.

## After this PR
==COMMIT_MSG==
Add a reason field to QosException
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/QosException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/QosException.java
@@ -34,13 +34,21 @@ import java.util.Optional;
  */
 public abstract class QosException extends RuntimeException {
 
+    private final QosReason reason;
+
     // Not meant for external subclassing.
-    private QosException(String message) {
+    private QosException(String message, QosReason reason) {
         super(message);
+        this.reason = reason;
     }
 
-    private QosException(String message, Throwable cause) {
+    private QosException(String message, Throwable cause, QosReason reason) {
         super(message, cause);
+        this.reason = reason;
+    }
+
+    public QosReason getReason() {
+        return reason;
     }
 
     public abstract <T> T accept(Visitor<T> visitor);
@@ -62,10 +70,24 @@ public abstract class QosException extends RuntimeException {
     }
 
     /**
+     * Like {@link #throttle()}, but includes a reason.
+     */
+    public static Throttle throttle(QosReason reason) {
+        return new Throttle(Optional.empty(), reason);
+    }
+
+    /**
      * Like {@link #throttle()}, but includes a cause.
      */
     public static Throttle throttle(Throwable cause) {
         return new Throttle(Optional.empty(), cause);
+    }
+
+    /**
+     * Like {@link #throttle()}, but includes a reason, and a cause.
+     */
+    public static Throttle throttle(QosReason reason, Throwable cause) {
+        return new Throttle(Optional.empty(), cause, reason);
     }
 
     /**
@@ -77,10 +99,24 @@ public abstract class QosException extends RuntimeException {
     }
 
     /**
+     * Like {@link #throttle(Duration)}, but includes a reason.
+     */
+    public static Throttle throttle(QosReason reason, Duration duration) {
+        return new Throttle(Optional.of(duration), reason);
+    }
+
+    /**
      * Like {@link #throttle(Duration)}, but includes a cause.
      */
     public static Throttle throttle(Duration duration, Throwable cause) {
         return new Throttle(Optional.of(duration), cause);
+    }
+
+    /**
+     * Like {@link #throttle(Duration)}, but includes a reason, and a cause.
+     */
+    public static Throttle throttle(QosReason reason, Duration duration, Throwable cause) {
+        return new Throttle(Optional.of(duration), cause, reason);
     }
 
     /**
@@ -92,10 +128,24 @@ public abstract class QosException extends RuntimeException {
     }
 
     /**
+     * Like {@link #retryOther(URL)}, but includes a reason.
+     */
+    public static RetryOther retryOther(QosReason reason, URL redirectTo) {
+        return new RetryOther(redirectTo, reason);
+    }
+
+    /**
      * Like {@link #retryOther(URL)}, but includes a cause.
      */
     public static RetryOther retryOther(URL redirectTo, Throwable cause) {
         return new RetryOther(redirectTo, cause);
+    }
+
+    /**
+     * Like {@link #retryOther(URL)}, but includes a reason, and a cause.
+     */
+    public static RetryOther retryOther(QosReason reason, URL redirectTo, Throwable cause) {
+        return new RetryOther(redirectTo, cause, reason);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.palantir.conjure.java.api.errors;
+
+import com.google.errorprone.annotations.CompileTimeConstant;
+import java.util.Objects;
+
+public final class QosReason {
+
+    @CompileTimeConstant
+    private final String name;
```
carterkozak, line 25: Let's rename the `name` field to `reason` for legibility.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private QosReason(@CompileTimeConstant String name) {
+        this.name = name;
+    }
+
+    public static QosReason of(@CompileTimeConstant String name) {
```
carterkozak, line 31: Should we validate the max-length and allowed characters here? Perhaps lower-case alphanumeric and hyphens? ~20 characters max? Plus some Javadoc to describe the constraints.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.api.errors;
+
+import com.google.errorprone.annotations.CompileTimeConstant;
+import java.util.Objects;
+
+public final class QosReason {
```
carterkozak, line 22: Let’s add some class level javadoc describing that the cardinality shouldn’t be terribly high on these

errors/src/test/java/com/palantir/conjure/java/api/errors/QosExceptionTest.java
```diff
@@ -47,4 +48,20 @@ public Class visit(QosException.Unavailable exception) {
...
+    @Test
+    public void testReason() {
+        QosReason reason = QosReason.of("reason");
+        assertThat(QosException.throttle(reason).getReason()).isEqualTo(reason);
+    }
+
+    @Test
+    public void testDefaultReasons() {
+        assertThat(QosException.throttle().getReason().toString()).isEqualTo("qos-throttle");
+        assertThatNoException().isThrownBy(() -> assertThat(QosException.retryOther(new URL("http://foo"))
```
carterkozak, line 61: `assertThatNoException().isThrownBy(() -> ` isn't terribly helpful here, if an exception is thrown, the test will fail

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
pkoenig10, line 22: What is the purpose of this class? It doesn't seem to provide any benefits over just using `String` directly.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosException.java
```diff
@@ -106,24 +156,52 @@ public static Unavailable unavailable() {
...
+    /**
+     * Like {@link #unavailable()}, but includes a reason, and a cause.
+     */
+    public static Unavailable unavailable(QosReason reason, Throwable cause) {
+        return new Unavailable(cause, reason);
+    }
+
     /** See {@link #throttle}. */
     public static final class Throttle extends QosException implements SafeLoggable {
+        private static final QosReason DEFAULT_REASON = QosReason.of("qos-throttle");
```
pkoenig10, line 182:
```suggestion
        private static final QosReason DEFAULT_REASON = QosReason.of("qos-throttle");

```

errors/src/main/java/com/palantir/conjure/java/api/errors/QosException.java
```diff
@@ -149,15 +227,26 @@ public List<Arg<?>> getArgs() {
 
     /** See {@link #retryOther}. */
     public static final class RetryOther extends QosException implements SafeLoggable {
+        private static final QosReason DEFAULT_REASON = QosReason.of("qos-retry-other");
```
pkoenig10, line 230:
```suggestion
        private static final QosReason DEFAULT_REASON = QosReason.of("qos-retry-other");

```

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
carterkozak: It should constrain the values to things that make sense as metric tags (disallow most special characters) and document cardinality. This is effectively an alias, we should document that QosReasons should be declared as static constants and reused rather than created per-qosexception (although cost-wise, `Throwable.fillInStackTrace` is far more expensive than creating new reason objects per-throw)
I prefer not to have raw-string or string+throwable constructors of private throwables because some libraries will attempt to recreate them unsafely/reflectively assuming the string is a message, which this is not.

errors/src/test/java/com/palantir/conjure/java/api/errors/QosExceptionTest.java
mpritham: I agree. I'm constructing a `URL` object, and need to catch the potential `MalformedUrlException` throw. I've decided to have the test throw the exception.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
carterkozak: Nice! Mind adding a SafeArg.of(“reason”, reason) so that we know the bad string when it does fail? Will make it a little easier for folks to debug.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -0,0 +1,74 @@
...
+ * throwing QosExceptions. The string used to construct a {@code Reason} object should be able to be used as a metric
+ * tag, for observability into {@link QosException} calls. As such, the string is constrained to have at most 20
+ * lowercase alphanumeric characters, and hyphens (-).
+ */
+public final class QosReason {
+
+    @CompileTimeConstant
+    private final String reason;
+
+    private static final Pattern REASON_REGEX = Pattern.compile("^[a-z0-9\\-]{1,20}$");
```
pkoenig10, line 38: 20 characters feels pretty limiting. Is there any concern with allowing slightly longer reasons - say 50 characters?

pkoenig10, line 38:
```suggestion
    private static final Pattern PATTERN = Pattern.compile("^[a-z0-9\\-]{1,20}$");
```

carterkozak: 50 sounds reasonable, I just want to avoid absurdly long message-style values in favor of keys that can be looked up.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
mpritham: Resolved this, but feel free to re-open @pkoenig10

mpritham:
Closing this PR. Opened #927 with the diff from this PR, on a branch of the target repository so that the changelog bot can run.

PR #927, opened 2022-12-12
Add a reason field to QosException

mpritham:
## Before this PR
Services throw `QosException`s to protect themselves from becoming overloaded. It's useful to know why a service responded with a `QosException` for better observability. However, the exceptions are currently logged at the `debug` level, which is disabled by default. It would be costly to log the exceptions at an `info` (or higher) level. 

This PR introduces new `QosException` factory methods which allow users to store a reason (an object of newly defined type `QosReason`) for throwing an exception.

## After this PR
==COMMIT_MSG==
Add a reason field to QosException
==COMMIT_MSG==

Note: This change has been reviewed in #925. This PR was created from a branch on the repository, instead of a fork, to enable the changelog bot.

### DIFF
--- a/changelog/@unreleased/pr-927.v2.yml
+++ b/changelog/@unreleased/pr-927.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add a reason field to QosException
+  links:
+  - https://github.com/palantir/conjure-java-runtime-api/pull/927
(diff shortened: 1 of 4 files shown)


carterkozak:
Good work, I'm happy with this. I'll plan to merge end of day unless other folks have concerns.

carterkozak (approved):
thanks!

PR #929, opened 2022-12-14
Ensure QosReason cannot be null

carterkozak:
## Before this PR
The API doesn't describe that nulls are allowed, however it didn't fail fast either when null values are provided.

## After this PR
==COMMIT_MSG==
Ensure QosReason cannot be provided as `null` by throwing an exception immediately
==COMMIT_MSG==


mpritham (approved):
Makes sense. 👍

PR #930, opened 2022-12-14
QosException reasons are included in SafeLoggable args

carterkozak:
## Before this PR
No insight into the qos reason unless the qos-exception was handled explicitly

## After this PR

==COMMIT_MSG==
QosException reasons are included in SafeLoggable args for better observability when exceptions are wrapped and logged.
==COMMIT_MSG==


mpritham (approved):
👍 Thank you

PR #932, opened 2022-12-15
Apply the Jetbrains Pattern annotation to the QosReason ctor

carterkozak:
This provides immediate IDE feedback when values are passed that don't satisfy our regex. This pattern is helpful for constant-string inputs, but requires additional IDE config if we begin to use it elsewhere that constants aren't required (e.g. ErrorType names) unless we disable the editor validation that all inputs can be traced to a constant value.
Screen Shot 2022-12-15 at 1 17 30 PM

==COMMIT_MSG==
Apply the Jetbrains Pattern annotation to the QosReason ctor
==COMMIT_MSG==

## Possible downsides?
API dependency on jetbrains annotations. We already do this in safe-logging preconditions and conjure for their `@Contract` annotation.

### DIFF
--- a/errors/build.gradle
+++ b/errors/build.gradle
@@ -7,6 +7,7 @@ dependencies {
     api "com.fasterxml.jackson.core:jackson-databind"
     api "com.google.code.findbugs:jsr305"
     api "com.palantir.safe-logging:safe-logging"
+    api "org.jetbrains:annotations"
     implementation "com.palantir.safe-logging:preconditions"
 
     testImplementation project(":extras:jackson-support")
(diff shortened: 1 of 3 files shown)


errors/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     api "com.fasterxml.jackson.core:jackson-databind"
     api "com.google.code.findbugs:jsr305"
     api "com.palantir.safe-logging:safe-logging"
+    api "org.jetbrains:annotations"
```
CRogers, line 10: Seems odd to actually ship this to prod - is it possible to use it as `compileOnly`?

CRogers: I guess it is in the API though... we'd probably want `compileOnly` for this dep in actual services, if such a thing was possible.

carterkozak: Agreed — it causes failures in transitives if the dep isn’t available when they compile, we really want apiCompileOnly or similar so that it’s exported…

CRogers:
:+1:

PR #983, opened 2023-04-20
Better error message for HostAndPort containing a protocol

ash211:
## Before this PR
There was an internal report (`ptcom/public-docs/pull/1359`) of a long debugging session that root caused to incorrectly specifying `hostAndPort: https://proxy-host.com:3128` instead of `hostAndPort: proxy-host.com:3128` in this configuration.

Before this PR, that would be easy to miss in the error message:

```
Given hostname does not contain a port number: {hostname=[http://squid:3128]}
```

## After this PR
==COMMIT_MSG==
Better error message for HostAndPort containing a protocol
==COMMIT_MSG==

With this PR, the error message would instead be:
```
hostPortString must not contain a protocol prefix like http:// or https://
```

To see the behavior change more clearly, look at just the second commit's diff (the first commit adds tests establishing current behavior).

## Possible downsides?
This PR performs more validation in the HostAndPort class, so any existing instances of values containing `://` would now start failing to deserialize.  The only usage of HostAndPort in this repo is in ProxyConfiguration which already fails during deserialization: https://github.com/palantir/conjure-java-runtime-api/blob/acf99a6a77403cd8080b42ce34fd298b766d01ca/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java#L95-L97.  So there would have to be a usage of HostAndPort directly, not via ProxyConfiguration.  Searching internal sourcegraph for `import com.palantir.conjure.java.api.config.service.HostAndPort` I see only one repo doing this.

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/HostAndPort.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/HostAndPort.java
@@ -34,6 +34,7 @@ package com.palantir.conjure.java.api.config.service;
 
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
 import org.jetbrains.annotations.VisibleForTesting;
 
 /** See Guava's {@code HostAndPort}. This class is a re-implementation that throws safelog exceptions. */
@@ -92,6 +93,16 @@ public final class HostAndPort {
             Preconditions.checkArgument(
                     isValidPort(port), "Port number out of range", SafeArg.of("port", hostPortString));
         }
+
+        // Additional validation for common errors. An alternative to this diff between conjure HostAndPort and
+        // Guava HostAndPort would be to use Guava's HostSpecifier.from() here instead.  But that would require
+        // either re-adding a Guava dependency (removed earlier in this repo's history), or copying that class
+        // and its dependencies (InetAddresses, InternetDomainName, Ascii.toLowerCase, Splitter, Joiner, and more)
+        // into this repo like we did with HostAndPort.
+        Preconditions.checkArgument(
+                !host.contains("://"),
+                "hostPortString must not contain a protocol prefix like http:// or https://",
+                UnsafeArg.of("hostPortString", hostPortString));
 
         return new HostAndPort(host, port);
     }
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/HostAndPort.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/HostAndPort.java
@@ -34,8 +34,9 @@ package com.palantir.conjure.java.api.config.service;
 
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
+import org.jetbrains.annotations.VisibleForTesting;
 
-/** See Guava's {@code HostAndPort}. */
+/** See Guava's {@code HostAndPort}. This class is a re-implementation that throws safelog exceptions. */
 public final class HostAndPort {
     /** Magic value indicating the absence of a port number. */
     private static final int NO_PORT = -1;
@@ -146,5 +147,15 @@ public final class HostAndPort {
     /** Return true for valid port numbers. */
     private static boolean isValidPort(int port) {
         return port >= 0 && port <= 65535;
+    }
+
+    @VisibleForTesting
+    String getHost() {
+        return host;
+    }
+
+    @VisibleForTesting
+    int getPort() {
+        return port;
     }
 }
--- a/service-config/src/test/java/com/palantir/conjure/java/api/config/service/HostAndPortTest.java
+++ b/service-config/src/test/java/com/palantir/conjure/java/api/config/service/HostAndPortTest.java
@@ -1,0 +1,38 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/HostAndPort.java
```diff
@@ -92,6 +94,16 @@ public static HostAndPort fromString(String hostPortString) {
...
 
+        // Additional validation for common errors. An alternative to this diff between conjure HostAndPort and
+        // Guava HostAndPort would be to use Guava's HostSpecifier.from() here instead.  But that would require
+        // either re-adding a Guava dependency (removed earlier in this repo's history), or copying that class
+        // and its dependencies (InetAddresses, InternetDomainName, Ascii.toLowerCase, Splitter, Joiner, and more)
+        // into this repo like we did with HostAndPort.
+        Preconditions.checkArgument(
+                !host.contains("://"),
+                "hostPortString must not contain a protocol prefix like http:// or https://",
+                UnsafeArg.of("hostPortString", hostPortString));
```
schlosna, line 105: - Should this implement [RFC 1123 hostname/IPv4/IPv6 validations](https://datatracker.ietf.org/doc/html/rfc1123#page-13) for `host`?
- Should we allow creation of `HostAndPort` from a valid [RFC 2396 URI](https://datatracker.ietf.org/doc/html/rfc2396) via `java.net.URI#create(String)` with something like the following (though I hate the exception as control flow)?
- Should the proxy configuration deserialization & checks actually handle this more gracefully?

https://github.com/palantir/conjure-java-runtime-api/blob/acf99a6a77403cd8080b42ce34fd298b766d01ca/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java#L89-L98

```
    @Nullable
    private static HostAndPort tryParseUri(String hostPortString) {
        try {
            URI uri = URI.create(hostPortString);
            return new HostAndPort(uri.getHost(), uri.getPort());
        } catch (RuntimeException e) {
            return null;
        }
    }
```

service-config/src/test/java/com/palantir/conjure/java/api/config/service/HostAndPortTest.java
```diff
@@ -0,0 +1,54 @@
...
+                .hasMessageContaining("hostPortString must not contain a protocol prefix like http:// or https://")
+                .containsArgs(UnsafeArg.of("hostPortString", "https://example.com:8080"));
+    }
+
+    @Test
+    void fromString_onHostAndPortWithAbcProtocolPrefix_doesNotThrow() {
+        assertThatLoggableExceptionThrownBy(() -> HostAndPort.fromString("abc://example.com:8080"))
+                .hasMessageContaining("hostPortString must not contain a protocol prefix like http:// or https://")
+                .containsArgs(UnsafeArg.of("hostPortString", "abc://example.com:8080"));
+    }
```
schlosna, line 53: nit: the test names `*_doesNotThrow()` don't seem to match expectation that this actually throws

service-config/src/test/java/com/palantir/conjure/java/api/config/service/ProxyConfigurationTests.java
```diff
@@ -147,4 +149,14 @@ public void serDe() throws Exception {
...
     }
+
+    @Test
+    public void incorrectlyFormattedHostAndPort() {
+        assertThatLoggableExceptionThrownBy(() -> ProxyConfiguration.builder()
+                        .hostAndPort("http://squid:3128")
+                        .type(ProxyConfiguration.Type.HTTP)
+                        .build())
+                .hasMessageContaining("hostPortString must not contain a protocol prefix like http:// or https://")
+                .containsArgs(UnsafeArg.of("hostPortString", "http://squid:3128"));
```
schlosna, line 160: should we add similar tests for these to `serDe`?

service-config/src/main/java/com/palantir/conjure/java/api/config/service/HostAndPort.java
ash211: 1. I think full standards-compliant host validation like RFC 1123 would be idea.  This split host+port pair ends up being sent into `InetSocketAddress.createUnresolved` at https://github.com/palantir/conjure-java-runtime/blob/33525f67c7acc31106ef4311f2fd2f9ef3ac92ae/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java#L173

So we would preferably match whatever validation that the JDK does when opening the socket.

2. that might be nice.  So instead of providing a better exception message for `hostAndPort: http://example.com:8080`, accept it as equivalent to `example.com:8080`

3. By handling deser and checks more gracefully, do you mean deserialize successfully but into an error state, rather than failing to deserialize?

schlosna: Making the deserialization more lenient to permit `host-and-port: https://your.proxy.host.example.com:1234` seems reasonable to me, but I'm not super close to all the consumers here and would defer to @carterkozak. I think we would still want to check and throw if one specified a URI scheme not in {`http`, `https`, `mesh`, `socks`} or if non-scheme/host/port elements of URI are specified (e.g. path, authority, etc.).

For 3, I was just referring to the `serDe()` test above and adding similar test cases that reads raw json (or yaml) with various valid & invalid `host-and-port:` URIs 

https://github.com/palantir/conjure-java-runtime-api/blob/31afb48b2c743d2489622100bb0f11c6648aa898/service-config/src/test/java/com/palantir/conjure/java/api/config/service/ProxyConfigurationTests.java#L137-L151

PR #1026, opened 2023-09-11
(Unknown)RemoteException implement ResponseStatus

schlosna:
## Before this PR
Consumers that handle `UnknownRemoteException` and `RemoteException` via `getStatus()` often require duplicative code paths because there is not a shared interface to retrieve the HTTP response codes.

## After this PR

==COMMIT_MSG==
(Unknown)RemoteException implement ResponseStatus
==COMMIT_MSG==

## Possible downsides?
We already expose `getStatus()` as public API on both exception types, but do we want to expose this interface as super type?

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/ResponseStatus.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/ResponseStatus.java
@@ -1,0 +1,22 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.api.errors;
+
+public interface ResponseStatus {
+    /** The HTTP status code of the HTTP response conveying the error. */
+    int getStatus();
+}
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
@@ -20,18 +20,17 @@ import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
 import com.palantir.logsafe.UnsafeArg;
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 
 /** An exception thrown by an RPC client to indicate remote/server-side failure from a non-remoting server. */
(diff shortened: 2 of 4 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/ResponseStatus.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.api.errors;
+
+public interface ResponseStatus {
+    /** The HTTP status code of the HTTP response conveying the error. */
+    int getStatus();
```
carterkozak, line 21: I don't have a strong opinion on this. Below is a light braindump thinking through it

- I'm not sure how much this buys us since most cases will still require us to check/cast to use the method.
- It's not entirely clear if other exceptions should support this sort of thing (QosException subtypes?).
- ServiceException is used within a server to produce a specific error, while RemoteException and UnknownRemoteException are thrown by clients to describe an error response. QosException sort of does both.
- Would supporting this interface create the impression that it covers all known response-related exceptions?

errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
```diff
@@ -54,9 +53,6 @@ public String getLogMessage() {
 
     @Override
     public List<Arg<?>> getArgs() {
-        List<Arg<?>> args = new ArrayList<>(2);
-        args.add(SafeArg.of("status", getStatus()));
-        args.add(UnsafeArg.of("body", getBody()));
-        return Collections.unmodifiableList(args);
+        return List.of(SafeArg.of("status", getStatus()), UnsafeArg.of("body", getBody()));
```
carterkozak, line 56: Any chance you could pull the String.format + collection creation improvements into a separate PR? Would love to get those improvements out, and merging doesn't require me to think as hard about api tradeoffs :-)

schlosna: no problem, split that out into https://github.com/palantir/conjure-java-runtime-api/pull/1027

errors/src/main/java/com/palantir/conjure/java/api/errors/ResponseStatus.java
schlosna: I'm on the fence here -- I think there are some cases where it is useful to expose the HTTP status code that triggered the `RemoteException`, though I think for this case ["When in doubt, leave it out"](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/32713.pdf). I'm leaning toward just closing this PR.

> * I'm not sure how much this buys us since most cases will still require us to check/cast to use the method.

The few places I was looking at where this would be useful would be something along the lines of below. I've refactored those places to just have 2 separate `catch` clauses that call the same method with the status code.

```
} catch (RemoteException | UnknownRemoteException e) {
    // logic using e.getStatus()
```

> * It's not entirely clear if other exceptions should support this sort of thing (QosException subtypes?).

I could see the status for `QosException` potentially being useful for some diagnostics, but probably not worth it and exposing those implementation details.

> * Would supporting this interface create the impression that it covers all known response-related exceptions?

I was thinking of this as something along the lines of [`SQLException#getErrorCode`](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/SQLException.html#getErrorCode()) though in this case they are [part of HTTP standards in RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-status-codes) and not vendor specific.

PR #1027, opened 2023-09-14
Remove String.format from exception creation

schlosna:
## After this PR

==COMMIT_MSG==
Remove String.format from exception creation
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
@@ -42,7 +42,7 @@ public final class UnknownRemoteException extends RuntimeException implements Sa
     }
 
     public UnknownRemoteException(int status, String body) {
-        super(String.format("Response status: %s", status));
+        super("Response status: " + status);
         this.status = status;
         this.body = body;
     }
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
@@ -19,6 +19,7 @@ package com.palantir.conjure.java.api.errors;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.Unsafe;
 import com.palantir.logsafe.UnsafeArg;
 import java.util.List;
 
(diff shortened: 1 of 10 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
```diff
@@ -42,7 +42,7 @@ public String getBody() {
     }
 
     public UnknownRemoteException(int status, String body) {
-        super(String.format("Response status: %s", status));
+        super("Response status: " + status);
```
carterkozak, line 45: Feel free to include the `List.of` optimization in this PR if you like

carterkozak (approved):
Thanks! Feel free to re-add merge-when-ready as is, I wanted to give you the option to add the args list optimization as well.

errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java
```diff
@@ -52,11 +51,9 @@ public String getLogMessage() {
...
     }
 
+    @Unsafe
     @Override
     public List<Arg<?>> getArgs() {
-        List<Arg<?>> args = new ArrayList<>(2);
-        args.add(SafeArg.of("status", getStatus()));
-        args.add(UnsafeArg.of("body", getBody()));
-        return Collections.unmodifiableList(args);
+        return List.of(SafeArg.of("status", getStatus()), UnsafeArg.of("body", getBody()));
```
schlosna, line 57: @carterkozak Note that changing this to using `List.of` triggered a `SafeLoggingPropagation` warning as the check could do deeper analysis, so I did a `-PerrorProneApply=SafeLoggingPropagation` as a separate commit. Let me know if you want to keep these as-is before merging or split into a separate follow-on PR.

```
> Task :errors:compileJava
/Volumes/git/palantir/conjure-java-runtime-api/errors/src/main/java/com/palantir/conjure/java/api/errors/UnknownRemoteException.java:54: Note: [SafeLoggingPropagation] Safe logging annotations should be propagated to encapsulating elements to allow static analysis tooling to work with as much information as possible. This check can be auto-fixed using `./gradlew classes testClasses -PerrorProneApply=SafeLoggingPropagation`
                        ^
    (see https://github.com/palantir/gradle-baseline#baseline-error-prone-checks)
  Did you mean '@Unsafe @Override'?
```

carterkozak: I think the changes outside of this file were handled by a recent baseline upgrade on develop (#1024), so the diff is pretty small in reality.

Unsafe annotations on `getArgs` look a bit off, but it's technically correct because it would be invalid to invoke something like `SafeArg.of("key", unknownRemoteException.getArgs())`. It's not something that we'd be likely to do, so the annotation probably isn't terribly valuable in this instance, but the pattern it's matching on is a safety unaware return type (`Arg` can be safe or unsage) where the method is known to return an `@Unsafe` type or value (in this case `UnsafeArg`).

PR #1056, opened 2023-11-09
Relax ErrorType namespace restrictions

pkoenig10:
## Before this PR
`ErrorType.create` cannot be used to create errors in the `Default` namespace.

This is problematic for endpoints that proxy requests and want to pass through any errors unchanged. We have a couple places internally with code that looks like this:

```java
try {
    // Proxy request
} catch (RemoteException exception) {
    SerializableError error = exception.getError();
    ErrorType.Code errorCode = ErrorType.Code.valueOf(error.errorCode());
    Arg<?>[] args = EntryStream.of(error.parameters())
            .mapKeyValue(UnsafeArg::of)
            .toArray(Arg[]::new);

    throw new ServiceException(ErrorType.create(errorCode, error.errorName()), exception, args);
}
```

I have recently realized that this is broken and the proxied requests end up failing with `Default:InvalidArgument` here  due to this restriction in `ErrorType.create`.

## After this PR
`ErrorType.create` can be used to create errors in the `Default` namespace.

## Possible downsides?

Is there some value to having this restriction?

### DIFF
--- a/changelog/@unreleased/pr-1056.v2.yml
+++ b/changelog/@unreleased/pr-1056.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: '`ErrorType.create` no longer prevents creating `ErrorType` values
+    in the `Default` namespace.'
+  links:
+  - https://github.com/palantir/conjure-java-runtime-api/pull/1056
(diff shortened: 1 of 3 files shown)


carterkozak (approved):
Agree that we should not throw exceptions in exceptional codepaths because it further deteriorates observability. We could return a sentinel value/replacement instead of allowing custom error creation in the default namespace, however we've never prevented folks from creating errortypes cross-namespace outside of the default (least helpful) namespace, so I don't see any reason to hold that restriction here either.

PR #1058, opened 2023-11-20
Make ServiceException and RemoteException assertions better

jdm2212:
==COMMIT_MSG==
When a ServiceException or RemoteException fails to satisfy test conditions, log the actual failing exception, including error instance ID.

This is useful for tracking down the error in the test container logs.

Errors now look like:
```
  "Expected error status to be 500, but found 501; remote exception: com.palantir.conjure.java.api.errors.RemoteException: RemoteException: FAILED_PRECONDITION (Default:FailedPrecondition) with instance ID a7507bd9-fcae-4cee-a888-f38710c30491
	at com.palantir.conjure.java.api.testing.RemoteExceptionAssertTest.lambda$testSanity$0(RemoteExceptionAssertTest.java:37)
	at org.assertj.core.api.ThrowableAssert.catchThrowable(ThrowableAssert.java:63)
	at org.assertj.core.api.AssertionsForClassTypes.catchThrowable(AssertionsForClassTypes.java:892)
	at org.assertj.core.api.Assertions.catchThrowable(Assertions.java:1366)
	at org.assertj.core.api.Assertions.assertThatThrownBy(Assertions.java:1210)
	at com.palantir.conjure.java.api.testing.RemoteExceptionAssertTest.testSanity(RemoteExceptionAssertTest.java:37)
```
==COMMIT_MSG==

### DIFF
--- a/test-utils/src/main/java/com/palantir/conjure/java/api/testing/RemoteExceptionAssert.java
+++ b/test-utils/src/main/java/com/palantir/conjure/java/api/testing/RemoteExceptionAssert.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.java.api.testing;
 
 import com.palantir.conjure.java.api.errors.ErrorType;
 import com.palantir.conjure.java.api.errors.RemoteException;
+import java.util.Objects;
 import org.assertj.core.api.AbstractThrowableAssert;
 
 public class RemoteExceptionAssert extends AbstractThrowableAssert<RemoteExceptionAssert, RemoteException> {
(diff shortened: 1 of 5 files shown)


test-utils/src/main/java/com/palantir/conjure/java/api/testing/ServiceExceptionAssert.java
```diff
@@ -58,15 +58,15 @@ public final ServiceExceptionAssert hasNoArgs() {
...
+            failWithMessage("Expected no args, but found %s; service exception: %s", allArgs, actual);
         }
 
         return this;
     }
 
-    private void failIfNotEqual(String message, Object expected, Object actual) {
-        if (!Objects.equals(expected, actual)) {
-            failWithMessage(message, expected, actual);
+    private <T> void failIfNotEqual(String message, T expectedValue, T actualValue) {
```
jdm2212, line 67: Had to change the `actual` parameter name here because it was hiding the `actual` field. Kinda surprised checkstyle or errorprone didn't flag that.

test-utils/src/main/java/com/palantir/conjure/java/api/testing/RemoteExceptionAssert.java
```diff
@@ -45,7 +45,7 @@ private <T> void failIfNotEqual(String fieldName, T expectedValue, T actualValue
         if (!Objects.equals(expectedValue, actualValue)) {
             failWithMessage(
                     "Expected %s to be %s, but found %s; remote exception: %s",
-                    fieldName, expectedValue, actualValue, actual);
+                    fieldName, expectedValue, actualValue, Assertions.printStackTrace(actual));
```
carterkozak, line 48: I think assertj escapes percents when the stack trace is concatenated into the format component, but I’m not sure it’s helpful if we add the stack trace as an arg.

I’m not sure why they add the trace into the format section, though, there must be a reason…

jdm2212: 🤦 removed the percent escaping

carterkozak (approved):
thanks!

PR #1085, opened 2024-01-22
Be more lenient when deserializing error parameters

LouisCaubet:
## Before this PR

`conjure-java` expects the error parameters to be a `Map<String, String>`, but `conjure-go-runtime` is serializing it as `Map<String, Object>`. If a conjure Go service throws an error with parameters that cannot be deserialized as `Map<String, String>`, deserialization will fail and the java service will throw an `UnknownRemoteException`.

## After this PR

We accept `Map<String, Object>` for parameters and keep stringified values in the deserialized `Map<String, String>`

==COMMIT_MSG==
Be more lenient when deserializing error parameters
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
@@ -87,7 +87,7 @@ public abstract class SerializableError implements Serializable {
     }
 
     /** A set of parameters that further explain the error. */
-    @JsonDeserialize(contentUsing = ParametersDeserializer.class)
+    @JsonDeserialize(contentUsing = ParameterDeserializer.class)
     public abstract Map<String, String> parameters();
 
     /**
@@ -136,7 +136,7 @@ public abstract class SerializableError implements Serializable {
         return new Builder();
     }
 
-    static class ParametersDeserializer extends JsonDeserializer<String> {
+    protected static final class ParameterDeserializer extends JsonDeserializer<String> {
 
         @Override
         public String deserialize(JsonParser parser, DeserializationContext _ctxt) throws IOException {
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
@@ -18,11 +18,18 @@ package com.palantir.conjure.java.api.errors;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.palantir.logsafe.Arg;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import java.io.IOException;
 import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -82,6 +89,7 @@ public abstract class SerializableError implements Serializable {
     }
 
     /** A set of parameters that further explain the error. */
+    @JsonDeserialize(using = ParametersDeserializer.class)
     public abstract Map<String, String> parameters();
 
     /**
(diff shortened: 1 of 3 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
```diff
@@ -82,6 +89,7 @@ public String errorInstanceId() {
     }
 
     /** A set of parameters that further explain the error. */
+    @JsonDeserialize(using = ParametersDeserializer.class)
```
carterkozak, line 92: I think we only want to define a map value deserializer here, right?

pkoenig10:
Are we sure we want to change the deserialization for all consumers of `SerializableError`? This feels a bit dangerous and potentially regrettable.

I was under the impression that we would only change the behavior in Dialogue by using a more lenient type here which we would then convert explicitly to `SerializableError`.

https://github.com/palantir/dialogue/blob/126ea2ba2d0fe4eb6e4de3b3e05772fe96ec75d1/dialogue-serde/src/main/java/com/palantir/conjure/java/dialogue/serde/ErrorDecoder.java#L125

errors/src/test/java/com/palantir/conjure/java/api/errors/SerializableErrorTest.java
```diff
@@ -193,6 +193,50 @@ public void testDeserializationFailsWhenNeitherErrorNameNorMessageIsSet() {
...
+                + "\"parameters\":{\"key\":true}}";
+        assertThat(deserialize(serialized))
+                .isEqualTo(SerializableError.builder()
+                        .from(ERROR)
+                        .putParameters("key", "true")
+                        .build());
+    }
+
+    @Test
+    public void testDeserializeWithJsonNumberParameters() throws Exception {
```
pkoenig10, line 230: For completeness, can we add a test with a string parameter?

errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
```diff
@@ -129,4 +134,12 @@ public static final class Builder extends ImmutableSerializableError.Builder {}
     public static Builder builder() {
         return new Builder();
     }
+
+    static class ParametersDeserializer extends JsonDeserializer<String> {
+
+        @Override
+        public String deserialize(JsonParser parser, DeserializationContext _ctxt) throws IOException {
+            return parser.readValueAsTree().toString();
```
pkoenig10, line 142: According to the documentation this will produce a JSON value, which isn't want we want here.

For example
```java
new TextNode("foo").toString();
```
produces
```java
"foo"
```
but we want
```
foo
```

---

Instead we want

```java
JsonNode node = parser.readValueAsTree();
String textValue = node.textValue();
if (textValue != null {
    return textValue;
} else {
    return node.toString();
}
```

carterkozak:
> Are we sure we want to change the deserialization for all consumers of SerializableError? This feels a bit dangerous and potentially regrettable.

The SerializableError type is meant to represent conjure errors, which it isn't doing quite right at the moment. Dialogue is our supported client, but I don't think we should fix this exclusively for dialogue consumers. In general I'd prefer to make error deserialization lax because we know things have already gone wrong, I don't think  making this specific deserializer more lenient will be regrettable. The java-side serializers are still constrained, which has the most potential wire API impact.

errors/src/test/java/com/palantir/conjure/java/api/errors/SerializableErrorTest.java
LouisCaubet: Oh right, I assumed that strings were tested by existing tests but turns out that wasn't the case.

errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
LouisCaubet: Good catch, fixed, thanks.

errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
```diff
@@ -129,4 +135,18 @@ public static final class Builder extends ImmutableSerializableError.Builder {}
     public static Builder builder() {
         return new Builder();
     }
+
+    static class ParametersDeserializer extends JsonDeserializer<String> {
```
pkoenig10, line 139:
```suggestion
    static class ParameterDeserializer extends JsonDeserializer<String> {
```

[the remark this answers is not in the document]
LouisCaubet: can't be made private because it needs to be accessible from `ImmutableSerializableError`, will make it protected.

pkoenig10:
One additional risk here is that clients are going to begin relying on this behavior are parsing out complex parameters from error responses. We will then be unable to change this behavior.

I don't think this is enough of a concern to block this PR, but I think it's worth calling out.

errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
```diff
@@ -129,4 +135,18 @@ public static final class Builder extends ImmutableSerializableError.Builder {}
     public static Builder builder() {
         return new Builder();
     }
+
+    protected static final class ParameterDeserializer extends JsonDeserializer<String> {
```
carterkozak, line 139: `protected` doesn't buy us anything over package-private here

LouisCaubet: so do you want me to remove it?

carterkozak: Yep, otherwise this change looks good to go!

carterkozak:
> One additional risk here is that clients are going to begin relying on this behavior are parsing out complex parameters from error responses. We will then be unable to change this behavior.
> 
> I don't think this is enough of a concern to block this PR, but I think it's worth calling out.

Agreed. We have the same issue on the serialization side (e.g. `OptionalInt[1]`, which I don't think we'll be able to escape. Either way, this much more intuitive behavior than failing to deserialize the failure and falling back to UnknownRemoteException.

errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
LouisCaubet: ok done!

carterkozak (approved):

PR #1103, opened 2024-02-13
Add ServiceExceptionAssert.hasCode

pkoenig10:
We have an internal library with methods that throw `ServiceException` with `ErrorType.PERMISSION_DENIED`. However , in order us flexibility to evolve this API, we don't guarantee that these methods will always use `ErrorType.PERMISSION_DENIED`, only that the error code will be `ErrorType.Code.PERMISSION_DENIED`. It would be nice if consumers of this library could make assertions about just the `ErrorType.Code` of a `ServiceException`.

Ultimately we may want something similar for `RemoteExceptionAssert`, but I don't have a concrete use case so it's unclear what the API should be given that `RemoteException` has both an `errorCode` and `status`.

### DIFF
--- a/changelog/@unreleased/pr-1103.v2.yml
+++ b/changelog/@unreleased/pr-1103.v2.yml
@@ -1,0 +1,6 @@
+type: feature
+feature:
+  description: Add `ServiceExceptionAssert.hasCode` to assert the `ErrorType.Code`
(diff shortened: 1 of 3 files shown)


carterkozak (approved):
Nice

PR #1158, opened 2024-04-19
ServiceException uses psuedorandom error instance UUIDs

schlosna:
## Before this PR

Remove contention on `NativePRNG` under high concurrency.

When some services are not ready, they throw `ServiceException`, leading to potential contention in `sun.security.provider.NativePRNG$RandomIO.implNextBytes(byte[])` 

image

## After this PR

==COMMIT_MSG==
ServiceException uses psuedorandom error instance UUIDs
==COMMIT_MSG==


carterkozak (approved):
Thanks!

PR #1166, opened 2024-05-10
Add support for HTTPS proxies

Eric-Alvarez:
Relates to https://github.com/palantir/dialogue/issues/2278
## Before this PR

The ProxyConfiguration API does not allow specifying an HTTPS proxy, a setup where the connection to the proxy is secured with HTTPS.
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==
The ProxyConfiguration API supports specifying HTTPS proxies. This requires changes in https://github.com/palantir/dialogue to honor this setting.


service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
```diff
@@ -92,6 +101,7 @@ protected final void check() {
         switch (type()) {
             case MESH:
             case HTTP:
+            case HTTPS:
```
carterkozak, line 104: I think line 135 needs to be updated as well

Eric-Alvarez: Fixed!

carterkozak (approved):

PR #1231, opened 2024-10-04
[WIP] Implement additional QoS metadata

carterkozak:
## Before this PR
No way to provide additional context with QoS exceptions.

## After this PR
==COMMIT_MSG==
Implement additional QoS metadata
==COMMIT_MSG==

This PR provides two pieces:
1. QosReason provides additional fields: `RetryHint` and `DueTo`
2. We provide a `QosReasons` utility which handles recording this metadata to a response, and recording from a response. This allows us to make changes/additions over time without updating projects that use these utilities.

## Possible downsides?
This adds complexity to our QoS exceptions, which are already somewhat complicated to reason about. We need additional control over how QoS responses impact the system as a whole, so this should be a reasonable trade-off.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
@@ -110,6 +110,8 @@ public final class QosReason {
 
         private Builder() {}
 
+        // other.reason must be a compile-time-constant
+        @SuppressWarnings("CompileTimeConstant")
         public Builder from(QosReason other) {
             return reason(other.reason()).retryHint(other.retryHint()).dueTo(other.dueTo());
         }
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
@@ -18,9 +18,13 @@ package com.palantir.conjure.java.api.errors;
 
 import com.google.errorprone.annotations.CompileTimeConstant;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.Safe;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.Objects;
-import java.util.regex.Pattern;
+import java.util.Optional;
+import javax.annotation.CheckReturnValue;
+import javax.annotation.Nullable;
 
 /**
  * A class representing the reason why a {@link QosException} was created.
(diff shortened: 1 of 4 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -61,16 +80,123 @@ public String toString() {
...
+    public enum RetryHint {
+        // TODO(ckozak): Should we declare the existing default value? Perhaps best if we only
+        // declare the minimal api to opt into new behavior.
+        //        /** Clients should attempt to retry this failure. */
+        //        EAGER,
+        /**
+         * Clients should not attempt to retry this failure,
+         * providing the failure as context back to the initial caller.
+         */
+        PROPAGATE;
```
bjlaub, line 162: minor nit: would something like "FAIL_FAST" be more descriptive? Errors will propagate no matter what, right? It's just a matter of how many retries happen in-between.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -61,16 +80,123 @@ public String toString() {
...
+        public Builder dueTo(Optional<DueTo> value) {
+            this.dueTo = Preconditions.checkNotNull(value, "dueTo").orElse(null);
+            return this;
+        }
+
+        public QosReason build() {
+            return new QosReason(
+                    Preconditions.checkNotNull(reason, "reason"),
+                    Optional.ofNullable(retryHint),
+                    Optional.ofNullable(dueTo));
```
bjlaub, line 149: it seems the presence of `retryHint` and `dueTo` imply some semantic meaning for the caller such that both must be set, and in fact [QosReasons#parseFromResponse](https://github.com/palantir/conjure-java-runtime-api/pull/1231/files#diff-4216445308ca6d6bea17e91cbe61d269a7ff3018ec8f2ccba5d383ef0cabd62cR39-R41) requires both to be set in order to parse either - but it's not clear from this API that you must set both. Should make that a requirement in the builder?

carterkozak: Yep, I have a table in the design document where the combinations are described with internal examples

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReasons.java
```diff
@@ -0,0 +1,100 @@
...
+
+import com.palantir.conjure.java.api.errors.QosReason.DueTo;
+import com.palantir.conjure.java.api.errors.QosReason.RetryHint;
+import java.util.Optional;
+
+public final class QosReasons {
+
+    private static final String CLIENT_REASON = "client-qos-response";
+    private static final QosReason DEFAULT_CLIENT_REASON = QosReason.of(CLIENT_REASON);
+    private static final String DUE_TO_HEADER = "Qos-Due-To";
```
bjlaub, line 27: do you want these to be all lowercase?

bjlaub (approved):
lgtm

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReasons.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final String RETRY_HINT_HEADER = "Qos-Retry-Hint";
+
+    private static final String DUE_TO_CUSTOM_HEADER_VALUE = "custom";
+    private static final String RETRY_HINT_DO_NOT_RETRY_HEADER_VALUE = "do-not-retry";
```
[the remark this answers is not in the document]
carterkozak: I like the separation of concerns with enum definitions in `QosReason.java` and the encoding pieces in the utility helper. Either way would be reasonable, but now that we have enum switch expressions, we can do a typesafe mapping with compile-time checks that we've associated a string with each enum value.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -61,16 +80,124 @@ public String toString() {
...
+     * {@link DueTo} informs relays this intent.
+     */
+    public enum DueTo {
+        /**
+         * A cause that the RPC system isn't directly aware of, for example a user or user-agent specific limit, or
+         * based on a specific resource that's being accessed, as opposed to the target node as a whole, or endpoint.
+         * QosReasons with this cause shouldn't impact things like the dialogue concurrency limiter.
+         */
+        CUSTOM;
+    }
```
carterkozak, line 179: I think we should use conjure-style enums here, with an `UNKNOWN` variant that can pass through new unknown values. This way, future additions will be passed through the system even before all services have taken the latest version of cjr-api. Additionally, it allows folks to potentially define more descriptive DueTo reasons (though it may require some care to make that safe, as it may impact dialogue and friends)

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
```diff
@@ -30,28 +34,43 @@
...
+        return reason;
+    }
+
+    public Optional<RetryHint> retryHint() {
+        return retryHint;
     }
 
+    public Optional<DueTo> dueTo() {
+        return dueTo;
+    }
```
carterkozak, line 71: I think this should become a `Set<DueTo> dueTo()` to support multiple reasons some day, keeps the door open.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
bjlaub: i think i'm struggling to understand the difference between `UNKNOWN` and `CUSTOM`. Is the concern that a client may receive a `DueTo` value from a newer version of c-j-r that they don't know how to deserialize?

carterkozak: > i think i'm struggling to understand the difference between `UNKNOWN` and `CUSTOM`.

I'm not entirely sure that it makes sense to declare both, in practice I think they could be considered equivalent by most consumers. I do prefer having well-defined values like `CUSTOM` for the common case, rather than using arbitrary string wrappers.
An open question I have is whether it would make sense to allow applications/libraries to define their own specific `DueTo` values using this mechanism -- such values would have to be treated the same as `CUSTOM` by Dialogue, but could provide better observability info.

> Is the concern that a client may receive a DueTo value from a newer version of c-j-r that they don't know how to deserialize?

Yep, the current implementation will ignore the value as though it doesn't exist.

errors/src/main/java/com/palantir/conjure/java/api/errors/QosReason.java
carterkozak: Hmm, a downside of allowing multiple `DueTo` values is that adding it to metric tags would become unwieldy.

bjlaub:
👍 lgtm

PR #1246, opened 2024-10-21
adding custom ConjureDefined exception types

mpritham:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java
```diff
@@ -129,6 +130,19 @@ public static SerializableError forException(ServiceException exception) {
         return builder.build();
     }
 
+    public static SerializableError forExceptionWithSerializedArgs(ServiceException exception) {
```
carterkozak, line 133: We don't necessarily need to reuse the `SerializableError` type, it's the serialized form as defined here that we want: https://palantir.github.io/conjure/#/docs/spec/wire?id=_55-conjure-errors

errors/src/main/java/com/palantir/conjure/java/api/errors/ConjureDefinedError.java
```diff
@@ -0,0 +1,16 @@
...
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeLoggable;
+
+// TODO(pm): according to java docs on Error an "Error - indicates serious problems that a reasonable application should
+//  not try to catch." Perhaps this isn't a good name, but we've already called something SerializableError and the
+//  "error" matches the conjure concept so this seems fine but check if folks care.
+// TODO(pm): copy-paste the ServiceException logic. I don't like extending ServiceException because we have
+//  (instanceof ServiceException) checks scattered across a few repos. Enumerate them, and see if they need to be
+//  handled differently compared to ServiceException.
+public final class ConjureDefinedError extends ServiceException implements SafeLoggable {
```
carterkozak, line 12: Shouldn't this be a checked exception? Extending `ServiceException` means we're also extending `RuntimeException`, which means it's unchecked.

PR #1262, opened 2024-11-12
Add checked service exception, used for endpoint associated errors

mpritham:
## Before this PR
In order to support work to generated checked exceptions for endpoint associated errors, we need to add a `CheckedServiceException`.

Please see [this PR in conjure](https://github.com/palantir/conjure/pull/1670) for high level context, and [this PR in `conjure-java`](https://github.com/palantir/conjure-java/pull/2401) for context on the Java-specific implementation.

## After this PR
==COMMIT_MSG==
Add a checked service exception, used for endpoint associated errors
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/CheckedServiceException.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/CheckedServiceException.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2017 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2024 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,32 +21,29 @@ import com.palantir.logsafe.SafeLoggable;
 import java.util.List;
 import javax.annotation.Nullable;
 
-/** A {@link ServiceException} thrown in server-side code to indicate server-side {@link ErrorType error states}. */
-public final class ServiceException extends RuntimeException implements SafeLoggable {
-    private static final String EXCEPTION_NAME = "ServiceException";
+/**
+ * An exception raised by a service to indicate an expected error state.
+ */
+public abstract class CheckedServiceException extends Exception implements SafeLoggable {
+    private static final String EXCEPTION_NAME = "CheckedServiceException";
     private final ErrorType errorType;
-    private final List<Arg<?>> args; // unmodifiable
-
+    private final List<Arg<?>> args; // This is an unmodifiable list.
     private final String errorInstanceId;
     private final String unsafeMessage;
     private final String noArgsMessage;
 
     /**
-     * Creates a new exception for the given error. All {@link com.palantir.logsafe.Arg parameters} are propagated to
-     * clients; they are serialized via {@link Object#toString}.
+     * Creates a new exception for the given error.
      */
-    public ServiceException(ErrorType errorType, Arg<?>... parameters) {
+    public CheckedServiceException(ErrorType errorType, Arg<?>... parameters) {
         this(errorType, null, parameters);
     }
 
     /** As above, but additionally records the cause of this exception. */
-    public ServiceException(ErrorType errorType, @Nullable Throwable cause, Arg<?>... args) {
-        // TODO(rfink): Memoize formatting?
+    public CheckedServiceException(ErrorType errorType, @Nullable Throwable cause, Arg<?>... args) {
         super(cause);
-
         this.errorInstanceId = ServiceExceptionUtils.generateErrorInstanceId(cause);
         this.errorType = errorType;
-        // Note that instantiators cannot mutate List<> args since it comes through copyToList in all code paths.
         this.args = ServiceExceptionUtils.arrayToUnmodifiableList(args);
         this.unsafeMessage = ServiceExceptionUtils.renderUnsafeMessage(EXCEPTION_NAME, errorType, args);
         this.noArgsMessage = ServiceExceptionUtils.renderNoArgsMessage(EXCEPTION_NAME, errorType);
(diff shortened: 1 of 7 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceError.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.api.errors;
+
+import com.palantir.logsafe.SafeLoggable;
+
+public interface ServiceError extends SafeLoggable {
```
mpritham, line 21: This is so I can use `getErrorType` in `AbstractServiceExceptionAssert` [(here)](https://github.com/palantir/conjure-java-runtime-api/pull/1262/files#diff-fab4bb1831ddc2f1ca2334e24f5c4ba290e7614f72bfa540f2690269568fe723R42).

This doesn't need to extend `SafeLoggable`, but I can't think of when we'd want to have a service error without support for arguments.

carterkozak: I'd prefer not to couple the two together via public API

mpritham: > I'd prefer not to couple the two together via public API

Yeah, I agree this feels clunky and exposes the ability for folks to use the interface directly which we don't want.

> the [safe-logging assertj assertThatLoggableException](https://github.com/palantir/safe-logging/blob/ba03dd3f1545c75495b7af81058b5eb04f766dad/preconditions-assertj/src/main/java/com/palantir/logsafe/testing/Assertions.java#L50-L59) can be used to test these using isInstanceOf/isInstanceOfSatisfying/hasArgs. It'll be a bit more involved to refactor, but I think the result is much easier to follow

To confirm, is the suggestion here to have test-writers use `assertThatLoggableExceptionIsThrown` in their tests, and writing those tests will be a bit more involved?

carterkozak: > To confirm, is the suggestion here to have test-writers use `assertThatLoggableExceptionThrownBy` in their tests, and writing those tests will be a bit more involved?

Yes, but I don't think writing those tests will actually be any more involved, only the migration to adopt checked exceptions if they previously used the assertions from this project.

errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceExceptionUtils.java
```diff
@@ -0,0 +1,130 @@
...
+    static String renderNoArgsMessage(String exceptionName, ErrorType errorType) {
+        return exceptionName + ": " + errorType.code() + " (" + errorType.name() + ")";
+    }
+
+    /**
+     * Finds the errorInstanceId of the most recent cause if present, otherwise generates a new random identifier. Note
+     * that this only searches {@link Throwable#getCause() causal exceptions}, not {@link Throwable#getSuppressed()
+     * suppressed causes}.
+     */
+    // VisibleForTesting
```
mpritham, line 106: I notice we don't use the `@VisibleForTesting` annotation in this repo.

errors/src/test/java/com/palantir/conjure/java/api/errors/CheckedServiceExceptionTest.java
```diff
@@ -0,0 +1,130 @@
...
+    @Test
+    public void testErrorIdsAreUnique() {
+        UUID errorId1 = UUID.fromString(new TestCheckedServiceException(ERROR).getErrorInstanceId());
+        UUID errorId2 = UUID.fromString(new TestCheckedServiceException(ERROR).getErrorInstanceId());
+
+        assertThat(errorId1).isNotEqualTo(errorId2);
+    }
+
+    @Test
+    public void testErrorIdsAreInheritedFromCheckedServiceExceptions() {
```
carterkozak, line 103: Let's add another variant of this test where the `rootCause` is a regular `ServiceException`, and the inverse in the `ServiceExceptionTest` where the `rootCause` is a `CheckedServiceException` for completeness

errors/src/main/java/com/palantir/conjure/java/api/errors/CheckedServiceException.java
```diff
@@ -0,0 +1,80 @@
...
+    private static final String EXCEPTION_NAME = "CheckedServiceException";
+    private final ErrorType errorType;
+    private final List<Arg<?>> args; // This is an unmodifiable list.
+    private final String errorInstanceId;
+    private final String unsafeMessage;
+    private final String noArgsMessage;
+
+    /**
+     * Creates a new exception for the given error.
+     */
```
carterkozak, line 37: let's duplicate the javadoc from ServiceException: https://github.com/palantir/conjure-java-runtime-api/blob/095d16b5101e3852c5d7c7451b044a5c3fd855df/errors/src/main/java/com/palantir/conjure/java/api/errors/ServiceException.java#L34-L37

carterkozak: (updated not to describe serialization)

carterkozak (approved):
A couple javadoc and test nits, approving to cut a release candidate based on this in the interim to unblock the other PRs :-]

errors/src/main/java/com/palantir/conjure/java/api/errors/CheckedServiceException.java
```diff
@@ -0,0 +1,81 @@
...
+        super(cause);
+        this.errorInstanceId = ServiceExceptionUtils.generateErrorInstanceId(cause);
+        this.errorType = errorType;
+        this.args = ServiceExceptionUtils.arrayToUnmodifiableList(args);
+        this.unsafeMessage = ServiceExceptionUtils.renderUnsafeMessage(EXCEPTION_NAME, errorType, args);
+        this.noArgsMessage = ServiceExceptionUtils.renderNoArgsMessage(EXCEPTION_NAME, errorType);
+    }
+
+    /** The {@link ErrorType} that gave rise to this exception. */
+    public ErrorType getErrorType() {
```
carterkozak, line 54: I'd expect this to fail compilation without being marked `final` -- the implemented methods should all be final so that subclasses cannot break the rules!

carterkozak: Maybe we've opted out of the checkstyle `DesignForExtension` check here?

errors/src/test/java/com/palantir/conjure/java/api/errors/TestCheckedServiceException.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.api.errors;
+
+import com.palantir.logsafe.Arg;
+
+final class TestCheckedServiceException extends CheckedServiceException {
```
bjlaub, line 21: minor nit: maybe call this `TestError` or something, the naming threw me off as it was so similar to `CheckedServiceExceptionTest`.

mpritham: Yeah that is confusing. Done.

bjlaub:
👍 this looks good to me

carterkozak (approved):
Thanks, @mpritham!

PR #1264, opened 2024-11-13
Minor update to doc string 

phegstrom:
## Before this PR

Doc referenced the message in ServiceException, however the serviceException.unsafeMessage or unsafeMessage.noArgsMessage is not referenced at all.

## After this PR

==COMMIT_MSG==
Correct doc string
==COMMIT_MSG==


carterkozak:
🤖

carterkozak (approved):

PR #1272, opened 2024-12-02
faster parsing for primary name component of user-agent strings

bjlaub:
## Before this PR
Some codepaths parse a user-agent string, but are only interested in the value of `UserAgents.tryParse(input).primary().name()`. Parsing the entire user-agent string can be somewhat expensive in this case, since it involves matching against multiple regular expressions, when in reality the caller is only interested in the string that immediately preceded the first `'/'` character in the string.

Note that there _is_ an existing fast-path for checking for a valid `name()` (which we re-use here) that avoids regular expressions, but the version number is still matched against a regex, and the entire parsing logic still relies on iterating over regex matches to pull out individual user-agent strings in the input, which is unnecessary in this case.

## After this PR
Adds a method `UserAgents#tryParsePrimaryName`, which attempts to parse and return _only_ the "name" component of the primary user-agent, and ignores everything else. It tries to be faster by explicitly not doing any regular expression matching and only examining a subset of the user-agent string (up to the very first "/" character).

==COMMIT_MSG==
faster parsing for primary name component of user-agent strings
==COMMIT_MSG==

## Possible downsides?
This is fairly specialized; it's debatable whether or not it belongs in the public API here, but it's helpful to re-use the fast-path that checks for valid name characters to ensure the result matches what `UserAgents.tryParse(input).primary().name()` would return.

Note that `UserAgents#tryParsePrimaryName` may have unintended results when given a user-agent string without any version number (e.g. the string `"foo/"`); it will return `foo`, even though `UserAgents#tryParse` would return `unknown/0.0.0` since it did not find a valid version number.

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
@@ -98,11 +98,7 @@ public final class UserAgents {
             return "unknown";
         }
         int split = userAgent.indexOf('/');
-        if (split < 0) {
-            return "unknown";
-        }
-        String primaryName = userAgent.substring(0, split);
-        if (primaryName.isEmpty()) {
+        if (split <= 0) {
             return "unknown";
         }
         // the regex-based logic will only match on blocks of the form "foo/1.2.3 (some extra stuff)", where the name
@@ -110,22 +106,22 @@ public final class UserAgents {
         // it will possibly ignore leading characters (even non-whitespace) up until the "foo/1.2.3" part
         // so we further split primaryName into _just_ the set of characters immediately preceding
         // the "/" for which isValidCharForName() returns true.
-        int lindex = primaryName.length() - 1;
+        int lindex = split - 1;
         while (lindex >= 0) {
-            if (!isValidCharForName(primaryName.charAt(lindex))) {
+            if (!isValidCharForName(userAgent.charAt(lindex))) {
                 break;
             }
             --lindex;
         }
         // lindex points to the first invalid character encountered walking backwards, which we want to exclude
-        primaryName = primaryName.substring(lindex + 1);
+        ++lindex;
         // the first character must match isAlpha()
-        if (!isAlpha(primaryName.charAt(0))) {
+        if (!isAlpha(userAgent.charAt(lindex))) {
             return "unknown";
         }
         // we can avoid an extra call to isValidName() now, since we have validated that the first character matches
         // isAlpha(), and every character thereafter matches isValidCharForName()
-        return primaryName;
+        return userAgent.substring(lindex, split);
     }
 
     private static UserAgent parseInternal(String userAgent, boolean lenient) {
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
@@ -119,9 +119,12 @@ public final class UserAgents {
         }
         // lindex points to the first invalid character encountered walking backwards, which we want to exclude
         primaryName = primaryName.substring(lindex + 1);
-        if (!isValidName(primaryName)) {
+        // the first character must match isAlpha()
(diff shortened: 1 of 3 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -86,6 +86,45 @@ public static UserAgent tryParse(String userAgent) {
...
+     */
+    public static String tryParsePrimaryName(String userAgent) {
+        if (userAgent == null) {
+            return "unknown";
+        }
+        int split = userAgent.indexOf('/');
+        if (split < 0) {
+            return "unknown";
+        }
+        String primaryName = userAgent.substring(0, split);
```
carterkozak, line 104: could we implement this with only a single `substring` invocation by updating the utility methods to take substring bounds (e.g. string, offset, length)?

bjlaub: i think with https://github.com/palantir/conjure-java-runtime-api/pull/1272/commits/2ede87aef385975c164ee2745c5a72ba484b9861 we can avoid this `substring` invocation, as from this point forward we are only looking at individual characters between `0` and `split`

service-config/src/test/java/com/palantir/conjure/java/api/config/service/UserAgentTest.java
```diff
@@ -190,12 +190,44 @@ public void tryParse_parsesWithBestEffort() {
...
+        assertThat(UserAgents.tryParsePrimaryName("foo bar baz serviceA/1.2.3 (some stuff)"))
+                .isEqualTo("serviceA");
+        assertThat(UserAgents.tryParsePrimaryName(" serviceA/1.2.3")).isEqualTo("serviceA");
+        assertThat(UserAgents.tryParsePrimaryName("\tserviceA/1.2.3")).isEqualTo("serviceA");
+
+        // tryParsePrimaryName explicitly tries to avoid doing extra work like parsing a version string, which
+        // relies on a regex match that could be expensive (and we're not interested in it anyway)
+        // this means calls can sometimes have unintended behavior for a user-agent string that's not well formed,
+        // but that's probably okay for most use cases
+        assertThat(UserAgents.tryParsePrimaryName("foo/")).isEqualTo("foo");
```
carterkozak, line 228: Let's break this out into a separate test case, and update the inputs above which use `UserAgents.tryParsePrimaryName` to use a test util that parses with both `tryParsePrimaryName` and `tryParse(agent).primary().name()` to assert equivalence.

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
carterkozak: Makes sense

> which utility methods are you referring to?

I was thinking `isValidName` could take offset+len and be reused for this validation

service-config/src/test/java/com/palantir/conjure/java/api/config/service/UserAgentTest.java
bjlaub: okay, fixed

bjlaub:
I wrote a benchmark to validate that this is faster than `UserAgents.tryParse(userAgent).primary().name()`, I'm not sure it's worth adding to this change but here are the results:
```
Benchmark                                                                              (userAgent)  Mode  Cnt  Score    Error  Units
UserAgentsParseBenchmark.tryParse                                                   serviceA/1.2.3  avgt    4  0.177 ±  0.009  us/op
UserAgentsParseBenchmark.tryParse             serviceA/1.2.3 (foo) serviceB/4.5.6 (this is a test)  avgt    4  0.661 ±  0.030  us/op
UserAgentsParseBenchmark.tryParsePrimaryName                                        serviceA/1.2.3  avgt    4  0.015 ±  0.001  us/op
UserAgentsParseBenchmark.tryParsePrimaryName  serviceA/1.2.3 (foo) serviceB/4.5.6 (this is a test)  avgt    4  0.015 ±  0.001  us/op
```

carterkozak (approved):

PR #1283, opened 2025-01-22
Optimize UserAgents

schlosna:
## Before this PR

UserAgents parsing and validation used unnecessary regex capturing groups and regex split.

## After this PR

==COMMIT_MSG==
Avoid regex capturing groups and regex split
==COMMIT_MSG==

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
@@ -45,10 +45,8 @@ public final class UserAgents {
             Pattern.compile("^[0-9]+(?:\\.[0-9]+)*(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?$");
     private static final Pattern SEGMENT_PATTERN =
             Pattern.compile(String.format("(%s)/(%s)( \\((.+?)\\))?", NAME_REGEX, LENIENT_VERSION_REGEX));
-    private static final Splitter COMMA_OR_SEMICOLON_SPLITTER = Splitter.on(
-                    CharMatcher.anyOf(",;").precomputed())
-            .omitEmptyStrings()
-            .trimResults();
+    private static final Splitter COMMA_OR_SEMICOLON_SPLITTER =
+            Splitter.on(CharMatcher.anyOf(",;").precomputed()).trimResults().omitEmptyStrings();
 
     private UserAgents() {}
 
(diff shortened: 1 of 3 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -175,7 +181,7 @@ private static UserAgent parseInternal(String userAgent, boolean lenient) {
 
     private static Map<String, String> parseComments(String commentsString) {
         Map<String, String> comments = new HashMap<>();
-        for (String comment : commentsString.split("[,;]")) {
```
schlosna, line 178: In addition to unnecessary regex, this warns:

```
/Volumes/git/palantir/conjure-java-runtime-api/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java:178: warning: [StringSplitter] String.split(String) has surprising behavior
        for (String comment : commentsString.split("[,;]")) {
                                                  ^
    (see https://errorprone.info/bugpattern/StringSplitter)
  Did you mean 'for (String comment : Splitter.onPattern("[,;]").split(commentsString)) {'?
```

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -40,9 +42,13 @@ public final class UserAgents {
...
     private static final Pattern NODE_REGEX = Pattern.compile("[a-zA-Z0-9][a-zA-Z0-9.\\-]*");
     private static final Pattern VERSION_REGEX =
-            Pattern.compile("^[0-9]+(\\.[0-9]+)*(-rc[0-9]+)?(-[0-9]+-g[a-f0-9]+)?$");
+            Pattern.compile("^[0-9]+(?:\\.[0-9]+)*(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?$");
     private static final Pattern SEGMENT_PATTERN =
             Pattern.compile(String.format("(%s)/(%s)( \\((.+?)\\))?", NAME_REGEX, LENIENT_VERSION_REGEX));
+    private static final Splitter COMMA_OR_SEMICOLON_SPLITTER = Splitter.on(
+                    CharMatcher.anyOf(",;").precomputed())
+            .omitEmptyStrings()
+            .trimResults();
```
carterkozak, line 51: Does the order of these two methods impact the results? I imagine a string trimmed from “   “ to “” should be omitted (I don’t recall if that happens in this order)

schlosna: The calling order does not matter as `omitEmptyStrings()` & `trimResults()` both return a new Splitter with all of the configured splitter behavior plus the new behavior.

Regardless, updating to `trimResults().omitEmptyStrings()` as that is clearer and autoformats onto one less line :)

carterkozak (approved):

PR #1286, opened 2025-01-30
Conjure User-Agents supports arbitrary comment data

carterkozak:
This allows us to track more specific context through the platform

==COMMIT_MSG==
Conjure User-Agents supports arbitrary comment data
==COMMIT_MSG==

I need to put some more thought into this. A few areas where I think we may want to do things differently:

* I've replaced `nodeId` with `comments`, which can take multiple arbitrary values. `nodeId` was set on the top-level UserAgent, not on a per-agent basis, so the new comments are set on the top-level type and encoded just after the primary agent as well. I think we should move the comments to `Agent` instead, allowing each component to provide context.
* I'm encoding the comments as a list of strings. In practice the user-agent rfc considers the comment a single value, but we should be more prescriptive. We could require key-value pairs, as I expect that will be how most of these comments are used (much like nodeId), however there are quite a few examples in the wild where segments don't map nicely to key/value pairs. Each segment may have its own preferred delimiters/sequences/etc.

In general, I like that we retain more of the original info from browser user-agents, which might not be possible if we force key/value pairs. We could bias the API toward key/value pairs in a way that doesn't exclude other strings (e.g. empty value to represent a single string) but that would make parsing more complex and risk mutations when we parse->re-encode (e.g. delimiter characters in the comment key getting split, and moving into the value)

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
@@ -16,10 +16,12 @@
 
 package com.palantir.conjure.java.api.config.service;
 
+import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 import org.immutables.value.Value;
 
 /**
@@ -34,13 +36,14 @@ public interface UserAgent {
     /** Identifies the node (e.g., IP address, container identifier, etc) on which this user agent was constructed. */
     @Value.Lazy
     default Optional<String> nodeId() {
-        if (comments().isEmpty()) {
+        if (primary().comments().isEmpty()) {
             // fast path to avoid stream overhead
             return Optional.empty();
         }
-        return comments().stream()
+        return primary().comments().stream()
                 .filter(item -> item.startsWith("nodeId:"))
                 .map(item -> item.substring(7).trim())
+                .filter(UserAgents::isValidNodeId)
                 .findFirst();
     }
 
@@ -53,13 +56,15 @@ public interface UserAgent {
      */
     List<Agent> informational();
 
-    List<String> comments();
-
     /** Creates a new {@link UserAgent} with the given {@link #primary} agent and originating node id. */
     static UserAgent of(Agent agent, String nodeId) {
+        UserAgents.checkNodeId(nodeId);
+        List<String> comments = Stream.concat(
+                        agent.comments().stream().filter(item -> !item.startsWith("nodeId:")),
+                        Stream.of("nodeId:" + nodeId))
+                .toList();
         return ImmutableUserAgent.builder()
-                .addComments("nodeId:" + nodeId)
-                .primary(agent)
+                .primary(Agent.of(agent.name(), agent.version(), comments))
                 .build();
     }
 
@@ -89,20 +94,6 @@ public interface UserAgent {
         return ImmutableUserAgent.builder().from(this).addInformational(agent).build();
     }
 
-    default UserAgent addComment(String comment) {
-        UserAgents.checkComment(comment);
-        return ImmutableUserAgent.builder().from(this).addComments(comment).build();
-    }
-
-    @Value.Check
-    default void check() {
-        if (nodeId().isPresent()) {
-            if (!UserAgents.isValidNodeId(nodeId().get())) {
-                throw new SafeIllegalArgumentException("Illegal node id format", SafeArg.of("nodeId", nodeId().get()));
-            }
-        }
-    }
-
     /** Specifies an agent that participates (client-side) in an RPC call in terms of its name and version. */
     @Value.Immutable
     @ImmutablesStyle
@@ -112,6 +103,8 @@ public interface UserAgent {
         String name();
 
         String version();
+
+        List<String> comments();
 
         @Value.Check
         default void check() {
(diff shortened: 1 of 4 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
```diff
@@ -44,9 +53,14 @@ public interface UserAgent {
      */
     List<Agent> informational();
 
+    List<String> comments();
```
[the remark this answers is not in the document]
carterkozak: Needs javadoc

carterkozak:
> if we go forward with key-value pairs then I'd propose we model it as a list of key-value pairs, and not as a map. I.e. don't require the keys to be unique. It's quite common for us to have the same "key" with different values in our internal requests.
> I don't have a strong opinion if we should make this key-value pairs or just strings though - we don't currently parse them, and is easy enough to parse <resource-type>:() if I want to parse all my resource identifiers, as long as we agree on delimiters (e.g. white-space).

I tend to avoid using maps in apis, for uniqueness constraints as well -- I like the flexibility of a simple list of strings instead of dealing with key/value pair delimiters which may not be sufficient for all types of key/value pair. Let's stick with `List<String>` and allow consumers to handle the data however best suits them.

> Note that for my use-case, I don't really care about the browser user-agents.

I'm not suggesting that the additional metadata is useful for your use case, but rather than by implementing a feature that is helpful for you, we may be able to more closely align with how user-agents are constructed and used more broadly. This gives me confidence that the feature will be supportable without causing issues regardless of whether/how it's used for this particular use-case.

> In fact, they're noise. I care about which resource is being accessed, not by which client or browser (which kind of indicates we're extending the usability of user-agent to maybe not what it was designed to do).

This is a fair caveat -- the use-case that we discussed for this metadata a few weeks ago did have better alignment, describing state/intent of a caller rather than what specifically they were attempting to access.

> In fact, we're currently using internally /, which in fact is not supported by the user-agent RFC. 🤦

That's not how I interpreted the rfc, they seem entirely valid (and common) within both the agent and comment components. See examples at https://useragents.io/

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgents.java
```diff
@@ -45,18 +45,20 @@ public final class UserAgents {
             Pattern.compile("^[0-9]+(?:\\.[0-9]+)*(?:-rc[0-9]+)?(?:-[0-9]+-g[a-f0-9]+)?$");
     private static final Pattern SEGMENT_PATTERN =
             Pattern.compile(String.format("(%s)/(%s)( \\((.+?)\\))?", NAME_REGEX, LENIENT_VERSION_REGEX));
-    private static final Splitter COMMA_OR_SEMICOLON_SPLITTER =
-            Splitter.on(CharMatcher.anyOf(",;").precomputed()).trimResults().omitEmptyStrings();
+    private static final Splitter SEMICOLON_SPLITTER =
+            Splitter.on(CharMatcher.is(';').precomputed()).trimResults().omitEmptyStrings();
```
carterkozak, line 49: In the most recent commit, I've updated this to no longer split on commas. I haven't found examples where the comma is used as a delimiter like semicolon, but it is used within the common string `KHTML, like Gecko`.
Everywhere that we encoded nodeId was a single component, and that wouldn't be impacted (though I don't think we actually use the nodeId component anywhere these days): `(nodeId:value)`

service-config/src/main/java/com/palantir/conjure/java/api/config/service/UserAgent.java
```diff
@@ -75,26 +95,28 @@ default UserAgent addAgent(Agent agent) {
...
         String version();
 
+        /**
+         * <a href="https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3>rfc7231 section-5.5.3</a> comment
+         * metadata (as described by
+         * <a href="https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6">rfc7230 section-3.2.6</a>)
+         * for additional diagnostic information. Note that this library provides a much stricter set of allowed
+         * characters within comments than the linked RFCs to reduce complexity.
+         */
+        List<@Safe String> comments();
```
bjlaub, line 118: do we need to validate comments in the `check()` method as well?

carterkozak: I don't think so, there are two factory methods, only one takes comments, and it validates them before passing args. This way we avoid unnecessary work in the common case.

bjlaub: sure, but the same could be said about `name` and `version`, no? should we just remove the `check()` method? i think this is more of a question of if anyone does/will use `ImmutableAgent.builder()` directly.

> This way we avoid unnecessary work in the common case.
what's the "common case" here? using the factory that does not accept comments?

carterkozak: > i think this is more of a question of if anyone does/will use ImmutableAgent.builder() directly.

Right, `ImmutableAgent` is package private, so it can't be used from outside of this module (assuming no package-clobbering hackery)

> what's the "common case" here? using the factory that does not accept comments?

Correct, the existing factory method that folks already use today, which does not accept comments.

bjlaub: > avoid unnecessary work

isn't this just a conditional on `comments().size()`?

carterkozak: Yep!

In general I'm a bit biased against immutables `@Check` methods as well, since they add a public method to the interfaces API which isn't meant to be called by consumers of the interface.

bjlaub (approved):

PR #1306, opened 2025-04-03
 Remove JMH plugin

mpritham:
## Before this PR
When running `./gradlew :service-config:jmhCompileGeneratedClasses`, we got:

```
> Task :service-config:jmhCompileGeneratedClasses FAILED
error: plug-in not found: ErrorProne
1 error
```

[This is a known error in the JMH plugin](https://github.com/melix/jmh-gradle-plugin/issues/248)

We don't need to use the JMH plugin.

## After this PR
==COMMIT_MSG==
Remove JMH plugin
==COMMIT_MSG==

## Possible downsides?
It'd be useful to write a small gradle plugin to handle this boilerplate across a few of our repos.


mpritham:
[For reference, this commit shows what a workaround that doesn't remove the plugin would look like. ](https://github.com/palantir/conjure-java-runtime-api/pull/1306/commits/cb2885bfee5021f2f015f2b0f2751996d524d3ce)

But I don't think there's a reason to use this plugin when we can run JMH benchmarks without it.

pkoenig10 (approved):

PR #1383, opened 2025-07-14
Add `AbstractServiceError` and extend `ServiceExceptionAssert` to support `EndpointServiceException`

mpritham:
Do not merge yet. Out of an abundance of caution, I'd like to test this with an RC to validate there are no hiccups with Conjure error handling.

## Before this PR
When #1352 was added, the test utilities to assert properties on the `EndpointServiceException` was not added.

## After this PR
This PR introduces a common abstract base class `AbstractServiceError` which both `ServiceException` and `EndpointServiceException` extend. It also modifies `ServiceExceptionAssert` to use `AbstractServiceError` instead of `ServiceException` in the template variable. 

==COMMIT_MSG==
Add `AbstractServiceError` and extend `ServiceExceptionAssert` to support `EndpointServiceException`
==COMMIT_MSG==

## Possible downsides?
We are making `AbstractServiceError` public so we can use it in the `ServiceExceptionAssert` which lives in a different package. Ideally, `AbstractServiceError` will not be public. A warning not to use the class was added to the docstring.

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/AbstractServiceError.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/AbstractServiceError.java
@@ -1,0 +1,30 @@
+/*
+ * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.api.errors;
+
+import com.palantir.logsafe.SafeLoggable;
+
+/**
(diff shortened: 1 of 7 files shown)


kunalrkak (dismissed):

errors/src/main/java/com/palantir/conjure/java/api/errors/AbstractServiceError.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.palantir.conjure.java.api.errors;
+
+import com.palantir.logsafe.SafeLoggable;
+
+/**
+ * This class should not be used directly. The parent class for ServiceException and EndpointServiceException.
+ */
+public abstract class AbstractServiceError extends RuntimeException implements SafeLoggable {
```
aldexis, line 24: nit: Could we name this `AbstractServiceException` since it's not an `Error`?
```suggestion
public abstract class AbstractServiceException extends RuntimeException implements SafeLoggable {
```

errors/src/main/java/com/palantir/conjure/java/api/errors/AbstractServiceError.java
```diff
@@ -0,0 +1,30 @@
...
+    public abstract ErrorType getErrorType();
+
+    protected AbstractServiceError(Throwable cause) {
+        super(cause);
+    }
```
aldexis, line 29: Can we not share more implementation than this between the two existing exceptions? Or is the goal to share the minimum to start?

aldexis (approved):
I'm fairly sure this will not cause ABI breaks (we might want to add revapi if we don't already have it?), however reverting this change (if we end up reconsidering) would be one

PR #1388, opened 2025-07-17
Expand `ProxyConfiguration#Type` to include `FROM_ENVIRONMENT_NO_PROXY_AWARE`

mpritham:
## Before this PR
When the `FROM_ENVIRONMENT` proxy type is selected, we do not respect the `no_proxy` environment variable. We would like to start respecting this variable while not breaking existing clients that use `FROM_ENVIRONMENT`. Requiring that clients explicitly select the behavior that respects `no_proxy` allows us to roll the change out safely.

Eventually, we can deprecate the `FROM_ENVIRONMENT` type and remove it.

## After this PR

==COMMIT_MSG==
Expand `ProxyConfiguration#Type` to include `FROM_ENVIRONMENT_NO_PROXY_AWARE`
==COMMIT_MSG==

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/ProxyConfiguration.java
@@ -49,6 +49,12 @@ public abstract class ProxyConfiguration {
         FROM_ENVIRONMENT,
 
         /**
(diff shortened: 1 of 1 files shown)


mpritham:
Closing. This change is no longer needed at present. We can revisit support for no-proxy later.

mpritham:
[Opened this issue to track](https://github.com/palantir/conjure-java-runtime/issues/3166)

PR #1407, opened 2025-08-03
Configurable connection pool reuse policy

acsmyth:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/service-config/src/main/java/com/palantir/conjure/java/api/config/service/PartialServiceConfiguration.java
+++ b/service-config/src/main/java/com/palantir/conjure/java/api/config/service/PartialServiceConfiguration.java
@@ -92,6 +92,10 @@ public interface PartialServiceConfiguration {
     @JsonAlias("proxy-configuration")
     Optional<ProxyConfiguration> proxyConfiguration();
 
+    /** Reuse policy for connection pooling. If absent, uses client implementation default. */
+    @JsonAlias("pool-reuse-policy")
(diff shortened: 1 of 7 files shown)


service-config/src/main/java/com/palantir/conjure/java/api/config/service/PartialServiceConfiguration.java
```diff
@@ -92,6 +92,10 @@ public interface PartialServiceConfiguration {
     @JsonAlias("proxy-configuration")
     Optional<ProxyConfiguration> proxyConfiguration();
 
+    /** Reuse policy for connection pooling. If absent, uses client implementation default. */
+    @JsonAlias("pool-reuse-policy")
+    Optional<PoolReusePolicy> poolReusePolicy();
```
pkoenig10, line 97: Connection pools are often managed by the underlying HTTP client implementation. It's not clear to me that all client implementations we will ever use will allow us to configure the connection pool in this way. Does the Apache client used by Dialogue support this? At they very least, I'd like to see functioning CJR and Dialogue PRs before merging this.

pkoenig10 (commented):
I'm not convinced that the cost of adding this configurability is worth the benefit.

If the default LIFO behavior of the Apache HTTP client is causing our clients to recreate connections, then it seems like these client simply aren't making enough requests where the connection creation matters.

We have existing server configuration options to control the `Keep-Alive` response header that will keep connection alive.

PR #1415, opened 2025-08-12
Add `AbstractSerializableError` and make `RemoteException` non-final

mpritham:
## Before this PR

As part of the effort to migrate the error serialization format of all Conjure errors to JSON in a safe way, I'd like to generate exceptions that are sub-classes of `RemoteException`. When Conjure-generated Dialogue clients throw these new exception subclasses, any existing code that catches and handles the `RemoteException` will continue to work as expected.

## After this PR

==COMMIT_MSG==
Add `AbstractSerializableError` and make `RemoteException` non-final
==COMMIT_MSG==


bjlaub (dismissed):
approving for RC

errors/src/main/java/com/palantir/conjure/java/api/errors/RemoteException.java
```diff
@@ -24,7 +24,7 @@
 import java.util.List;
 
 /** An exception thrown by an RPC client to indicate remote/server-side failure. */
```
aldexis, line 26: Should probably include a description of why this is not final and that we don't expect users to manually subclass these

errors/src/main/java/com/palantir/conjure/java/api/errors/ConjureErrorParameterFormat.java
```diff
@@ -33,7 +38,10 @@
 @Safe
 public final class ConjureErrorParameterFormat {
     private static final String JSON = "JSON";
+    private static final String JSON_AND_JAVA_STRING = "JSON+JAVA-STRING";
```
aldexis, line 41: Sounds like we decided that this was not worth adding? If so, let's not forget to remove it

PR #1442, opened 2025-09-17
Add `parameterMap` to `AbstractSerializableError`

mpritham:
## Before this PR
Currently, the rich parameter types are only accessible for clients that are aware of the type of the parameters. There is a specific proxy-like use case where clients are not aware of the type of the parameters, and would like to access the parameter map in order to pass it along to a client that does have knowledge of the parameter types.

## After this PR

==COMMIT_MSG==
Add `parameterMap` to `AbstractSerializableError`
==COMMIT_MSG==

### DIFF
--- a/errors/src/main/java/com/palantir/conjure/java/api/errors/AbstractSerializableError.java
+++ b/errors/src/main/java/com/palantir/conjure/java/api/errors/AbstractSerializableError.java
@@ -16,6 +16,8 @@
 
 package com.palantir.conjure.java.api.errors;
 
+import java.util.Map;
+
 /**
  * Base class for serializable Conjure errors with generic parameter type support.
  * <p>
@@ -43,6 +45,7 @@ public abstract class AbstractSerializableError<T> {
     private final String errorName;
     private final String errorInstanceId;
     private final T parameters;
+    private final Map<String, Object> parameterMap;
 
     public final String errorCode() {
         return errorCode;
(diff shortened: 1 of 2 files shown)


errors/src/main/java/com/palantir/conjure/java/api/errors/AbstractSerializableError.java
```diff
@@ -60,10 +63,26 @@ public final T parameters() {
...
-    protected AbstractSerializableError(String errorCode, String errorName, String errorInstanceId, T parameters) {
+    /**
+     * A map of parameter names to their values. Ideally, {@code parameters()} should be used if the specific type
+     * {@code T} is known. For clients that do not know (and do not need to know) the specific type {@code T}, this map
+     * provides a way to access the error parameters. For example, a client may want to access the parameters for any
+     * error and serialize them in a custom format.
+     */
+    public final Map<String, Object> parameterMap() {
+        return parameterMap;
+    }
```
aldexis, line 74: Mentioned this in an internal thread, but I would rather we don't actually do this in the end. This map itself should be constructible with a jackson mapper afaik, and even then, our use-case for this is purely to forward the error itself, so I don't think they actually need the map, and could just pass in the parameters object itself for serialization?

The fact that this loses the safety annotations of the parameter keys (a benefit from using a rich class here) also drives me away from this approach

